PR #16, opened 2016-07-03
Added decklist sort by Faction then Set

CoryOwens:
I have my Netrunner collection in binders, each faction having its own binder, then cards entered in order of release. I have used the browser console to add a matching sorting option, but I thought it might be a good feature for the site, so... here you go. 

As a sidenote, I must say I'm really impressed with the flexibility of your code here. It was so incredibly easy to add this option.


alsciende:
Thanks for the kind words =)

alsciende:
Change deployed. Thank you for the PR!

alsciende:
image

CoryOwens:
I was just thinking, it might be nice to add the set symbol and card number next to the card name when sorted this way. It might also be good when sorting by Set alone.
When I get home, I might look at implementing that, if you are okay with that idea. I don't want to clutter up your UI too much.

alsciende:
I think it's worth trying, yes. When sorted by Set, the card name is followed by its number, like 

> 2x Snare! (#70)

The # could be replaced by the set symbol, and this format could be re-used when sorting "by Faction then Set".

PR #69, opened 2016-11-02
Add less than and greater than for cycle number

Smashman:


Smashman:
Thanks!

Let me know when it's live on the site please.

alsciende:
It's live. Thanks for the PR :)

Issue #110, opened 2017-04-28
The Proffesor is counting wrongly new Universal Influence


Reznorcito:
I just added a "Cerberus Lady H1" to an Professor deck and it is counting it as 1 influence spent when it is the first copy of the program so that influence shouldn't be counting at all.

alsciende:
That's because Cerberus Lady H1 is on the MWL. Each copy costs 1 extra influence, regardless of the identity's faction.

gereons:
That's how I understood the new system as well, but now Michael Boggs has stated on RLC that that's not what should happen - the professor completely ignores the influence for the first program he includes.

yads:
This is now fixed.

plural:
while I don't think a lot of people are building against MWL 1.2 , this doesn't actually appear to be fixed

![Screenshot 2019-05-12 at 4 32 52 PM](https://user-images.githubusercontent.com/396562/57588026-afed5980-74d3-11e9-9cc0-50a304a1096a.png)

PR #116, opened 2017-06-14
Update The Professor to work with MWL 1.2

yads:
Fix for Issue #110


yads:
@Alsciende I just realized there's a bug with this implementation, give me a sec to fix it up.

PR #121, opened 2017-07-11
Destroy every other qtip before creating a new one

### LINKED ISSUE #119: On-hover card preview doesn't display art
Using Chrome Version 59.0.3071.115 (Official Build) (64-bit), also checked in private mode (all addons disabled).
Screenshot of bug in action:
selection_023

Sometimes the art does show (for example, in the first preview generated on every page but also at random).

eljenso:
With this PR, all qtips will be removed from the DOM before creating any new.
Tried setting hidden and blur events for qtips (http://qtip2.com/options#events), but did not work well.
Closes #119.


alsciende (approved):

alsciende:
Thank you! Your PR is deployed.

PR #127, opened 2017-08-03
Fix adding comments on review page

pminten:
Currently the add comment link/button on the reviews page doesn't work because the relevant javascript is in zoom.js and only loaded when viewing a card's page. This bug manifests as the user's browser going to https://netrunnerdb.com/review/comment and getting a "Your comment is empty" error.

By adding zoom.js to the reviews page this is fixed. It does add some functions that are not needed for the reviews list so I'm not sure this is the optimal patch.


alsciende:
Thank you!

PR #131, opened 2017-08-14
Cache refresh check

stoppableforce:
Implements a Cache Refresh check that works similarly to the 1.1.1.1 ('onesies') format check.


alsciende (approved):

alsciende:
Thank you a lot, @stoppableforce

Issue #138, opened 2017-09-15
PAD Campaign from Revised Core Set does not satisfy alliance requirements for PAD Factory


MrCowbird:
Sorry for the triviality of this; I know PAD Factory is a junk card.

When using Revised Core Set PAD Campaigns, the alliance requirement is not considered fulfilled for the purposes of PAD Factory.

PR #142, opened 2017-09-20
update influence computation for PAD Factory

yads:
* include Core 2.0 PAD Campaign

### DIFF
--- a/src/AppBundle/Resources/public/js/nrdb.js
+++ b/src/AppBundle/Resources/public/js/nrdb.js
@@ -566,7 +566,7 @@ function get_influence_cost_of_card_in_deck(card) {
             }
         } else if (card.code === '10038') {
             // PAD Factory: 3 PAD Campaign
-            if (count_card_copies(NRDB.data.cards.find({indeck: {'$gt': 0}, code: '01109'})) === 3) {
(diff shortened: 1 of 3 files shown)


yads:
Fix for #138

src/AppBundle/Resources/public/js/nrdb.js
```diff
@@ -566,7 +566,7 @@ function get_influence_cost_of_card_in_deck(card) {
             }
         } else if (card.code === '10038') {
             // PAD Factory: 3 PAD Campaign
-            if (count_card_copies(NRDB.data.cards.find({indeck: {'$gt': 0}, code: '01109'})) === 3) {
+            if (count_card_copies(NRDB.data.cards.find({indeck: {'$gt': 0}, code: '01109'})) === 3 || count_card_copies(NRDB.data.cards.find({indeck: {'$gt': 0}, code: '20128'})) === 3) {
```
alsciende: You can use `NRDB.data.cards.find({indeck: {'$gt': 0}, code: ['01109','20128']})`. That way you'll only do 1 query to the db and count both Pad Campaign together.

yads: I was wondering if there was a shortcut like that. Thanks, I'll update and push up.

alsciende:
Thank you!

PR #145, opened 2017-09-28
Set MWL Code and Copy original notes when duplicating decks.

### LINKED ISSUE #71: Copying decklists doesn't set Legality
When copying a published decklist to my own decks, the copy is always marked as "Casual", even if the source had "MWL 1.1" set. It'd be great if the legality was copied as well.

plural:
Fixes #71


alsciende:
Thank you :)

PR #152, opened 2017-10-04
Add Hashing Value to URL Routing

jdeng89:
I honestly don't have any idea what I'm doing but here's my idea =|.

### DIFF
--- a/src/AppBundle/Controller/BuilderController.php
+++ b/src/AppBundle/Controller/BuilderController.php
@@ -706,9 +706,12 @@ class BuilderController extends Controller
     
     }
 
-    public function viewAction ($deck_id)
+    public function viewAction ($deck_id,$hash)
     {
-
+		if(!$hash) {
+			throw $this->createNotFoundException("Deck not shared");
+		}
+	
         $dbh = $this->get('doctrine')->getConnection();
         $rows = $dbh->executeQuery("SELECT
 				d.id,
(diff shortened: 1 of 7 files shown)


alsciende:
Thank you @jdeng89 

My thoughts: it's still possible and quite easy to find the hash, since there is no salt. An attacker knows the targeted id and only has to compute some hashes with a few dates around the probable creation date. Also, the hash computation needs to be put in a separate service, and the hash can be computed server-side and sent to the front-side, no need to compute it front-side.

alsciende:
So, the PR is interesting and a nice way to implement it. Create a service, inject %secret%, return hash based on deck.id, deck.creationDate and %secret%, and it should be good.

gereons:
Since I like to be prepared: have you already decided if the proposed URL scheme going to stay?

jdeng89:
Alright, I've removed the date (because I actually think there's a huge bug in it that I'm not sure how to fix) and I've added a salt instead which will randomize it per deck. (this is better than secret as it's random per deck)

jdeng89:
I didn't actually do this because I'm not sure what the routing behavior is when there isn't a third path (hash), but in order to keep old decklist URLs safe, I should be able to just check whether or not either a salt exists or is empty string, and allow sharing to occur if it's empty (since all new Decks should have the salt set).

NoahTheDuke:
Closing for now as the work needed to fix would be best suited for a new PR.

Issue #150, opened 2017-10-03
Card Reviews


ghost:
When a card review is done, this is usually based on a specific card pool. With Rotation coming and the new core I think it would be useful to track the card pool a review was valid for. The easiest would be if the one writing the review would select it, or if it's done automatically based on the date.

alsciende:
Something like "Written before First Rotation"?

ghost:
I was more thinking about the issue that the review is aligned to a status of the card game. So it would be interesting to know which cards where available at that point in time. Assuming ANR is successful we will have more rotations to come. And on top it might be interesting for someone reading it, what was the available data pack 2 years ago. So I thought it would be an improvement if we could track that along with the review.

jdcowf:
If we could add a "Most recent release at time of publication" to the review that would be a great way to date them, similar to how the "Packs up to" is used in deck viewer.

ghost:
Yes, this sounds like a good solution

ghost:
thanks for implementing it! Really like it.

ghost:
One thing I realized when updating the review, shouldn't it also update the latest pack at time of writing?

x3N1GM4x:
Tricky - what if someone comes back to one of their reviews a few months later and just corrects a typo or rephrases something? Would you want the "latest pack" to update, artificially making the review seem newer than it really is?

ghost:
Agree. Perhaps it's best to write a total new review.

PR #159, opened 2017-10-09
Add "latest pack at time of writing" to card reviews

avandever:
Basically I would like to start to become familiar with this codebase, so I grabbed what seemed like a easy-ish issue to start with, choosing #150  as my victim. After this change, for each Review we will perform a SQL query to get the name of the most recently released pack at the time of the review creation, and pass that along in the review data. I'm attaching a screenshot of how it renders from my test instance. I also touched a couple of lines just to make their indentation match their neighbors more closely.
capture

One yellow flag I noticed that is confusing me a bit, is that when I load the same page including this change it reports 80 db queries, whereas on master it only reports 11. Let me know if you'd like me to gather more info related to that.

### DIFF
--- a/src/AppBundle/Controller/SearchController.php
+++ b/src/AppBundle/Controller/SearchController.php
@@ -315,7 +315,6 @@ class SearchController extends Controller
 				}
 				$cardinfo['available'] = $availability[$pack->getCode()];
 				if($view == "zoom") {
-					$allsetsdata = $this->get('cards_data')->allsetsdata();
 						$cardinfo['reviews'] = $this->get('cards_data')->get_reviews($card);
 						$sets = $this->get('cards_data')->allsetsnocycledata();
 						foreach ($cardinfo['reviews'] as $key => $review) {
--- a/src/AppBundle/Controller/SearchController.php
+++ b/src/AppBundle/Controller/SearchController.php
@@ -316,19 +316,6 @@ class SearchController extends Controller
 				$cardinfo['available'] = $availability[$pack->getCode()];
 				if($view == "zoom") {
 					$cardinfo['reviews'] = $this->get('cards_data')->get_reviews($card);
-					$dbh = $this->get('doctrine')->getConnection();
-					foreach ($cardinfo['reviews'] as $key => $review) {
-						// Trouvez le dernier pack publié au moment de la création de l'avis
-						$rows = $dbh->executeQuery(
-							"SELECT name
-							FROM pack
-							WHERE date_release > ?
-							ORDER by date_release DESC
-							LIMIT 1",
-							array($review['date_creation']->format('U'))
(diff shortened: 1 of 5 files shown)


alsciende:
Thank you @avandever . I like the idea and the way you're displaying it in the page. I'm worried about the server code though, because too many sql queries is really bad for the performance.

I'd like to suggest an alternate algorithm that doesn't query the database and is fast enough. CardsData->allsetsnocycledata() returns a list of all packs ordered by date. For each review, iterate over the list of packs until you find one whose release date is anterior to the review, and add its name to your rulings data.

alsciende:
Another thing: I think the addition of "latest_pack" to the rulings data could be done in `CardsData.php`.

avandever:
Thanks for the feedback! I'll try to get this updated later this week.

avandever:
This is using the API you recommended, which I believe should be hitting the json repo rather than the db, but I'm still showing 80 db queries instead of 11. 51 of them are pulling rows out of the card table (filtering on pack), 16 from pack (filtering on cycle), and 2 are pulling from card but filtering on position as well as pack. Is it possible there's just something weird I'm missing?

avandever:
This latest version goes back to the original, but removes the never-used call to CardsData->allsetsdata. And now the query count has gone up by only 1, from 11 to 12, which is much better than the 63 I was getting using allsetsnocycledata, I think.

Please let me know what you think. Sorry for all the churn. I'm once again reminded that my brain doesn't work so well after 11pm. :)
capture

alsciende:
I fixed (1453b082e5f5bf87d36bdcd88f1e96a55d2fac15) the issue with CardsData->allsetsdata() triggering a large number of queries. 

Please use it for your computation, and only modify CardsData->get_reviews(), not SearchController.

alsciende:
> This is using the API you recommended, which I believe should be hitting the json repo rather than the db

Prod doesn't read the json repo, never :) A db read is much much faster, in all cases.

avandever:
Is this where you imagine it going? This definitely seems like a better place. As far as db load though, I'm still getting 64 queries in 49.51 ms versus 12 queries in 7.45 ms if I go back to be7b795. Again, it's very possible I'm missing something obvious, but I did double-check that 
a107252 is applied within my repo.

alsciende:
You're right, there were still a high number of queries because I fixed allsetsdata() and you used allsetsnocycledata(). Anyway, none of these methods was the right choice since you only needed the packs. I ended up doing a simple findBy() on the Packs repository.

alsciende:
It's live :)

avandever:
Awesome! Thanks for the pointers!

avandever:
Oh geez, reminds me of http://www.snopes.com/business/genius/where.asp

Sent from Mail for Windows 10

alsciende:
deployed!

PR #167, opened 2017-11-26
fix error 500 when duplicating a deck without mwl info

### LINKED ISSUE #155: Error when trying to duplicate a decklist
In the "My Decks" section, when I select a deck in the list and hit the yellow Duplicate button, I'm thrown to a "500 Internal Server Error", with a 503 Service Unavailable at the top. It does this for any deck that I attempt to hit the duplicate button on.

gereons:
This should fix #155


alsciende:
So that's where the bug was! Thank you @gereons .

plural:
thank you!  Sorry for causing this issue with my previous CL.

PR #173, opened 2017-12-09
fix setting MWL, in particular to "Casual"

gereons:
This is a fix for the issue described in https://www.reddit.com/r/Netrunner/comments/7i9hlr/netrunnerdb_is_not_letting_me_switch_deck/ - previously it was impossible to reset a deck from any MWL version back to casual. Also and adding to the confusion, the MWL dropdown did not show "Casual" when the deck was in fact for casual play.


alsciende:
Deployed, thanks

PR #183, opened 2017-12-29
show unicorn for MWL restricted cards

gereons:


alsciende:
Unicorn, really?

gereons:
I added this in Net Deck and Jnet followed suit, so I thought it was only fitting :)

PR #192, opened 2018-02-01
migrate to Symfony 3.4 app structure

alsciende:
This PR moves a lot of files around to adopt the recommended application structure for Symfony 3.4. Also I improved the formatting of all php files. I'm going to test it a bit more before merging, but it should hit prod soon.


alsciende:
@gereons FYI

gereons:
Thanks for the heads-up!

PR #203, opened 2018-03-02
remove stray "+"

gereons:


gereons:
Also (accidentally commited here), use https for image urls to reduce the number of "mixed content" warnings.

alsciende:
nice!

PR #207, opened 2018-03-09
Checks and fixes for Rotation

gereons:
This PR attempts to fix several problems with creating and publishing post-rotation valid decks.

- it adds a "Rotation compliance" check to the deckbuilder, selectable in Settings.
- it adds a "Convert to post-rotation" to the warning message
- it removes rotated packs from the default values of the Collections tab


web/js/nrdb.js
```diff
@@ -709,6 +714,61 @@ function check_deck_limit() {
...
+        	});
+        	NRDB.data.cards.updateById(oldCode, {
+        		indeck : 0
+        	});
+            ++replaced;
+            Deck_changed_since_last_autosave = true;
+        }
+    });
+    
+    alert("Replaced " + replaced + " card(s) with their post-rotation counterparts.");
```
alsciende: i'd rather have this message appear in #rotated

alsciende (changes requested):

alsciende:
I wrote the review days ago but didn't validate it. You should be able to see the requested changes now.

web/js/nrdb.js
gereons: That's kinda hard, since `#rotated` gets emptied and hidden immediately after the conversion, in `update_deck()` which I think is necessary to call.

alsciende: Can't we call update_deck() first, then update `#rotated`? User clicks the "replace" link, link is emptied and immediately replaced with "X cards replaced".

gereons: Sure, why didn't I think of that? ;)

alsciende:
Thanks! I'm going to test it a bit; I'll let you know when the server is updated.

PR #209, opened 2018-03-12
Invert rotation logic

gereons:
Implements the data change from https://github.com/Alsciende/netrunner-cards-json/pull/326.

When this gets merged, the steps to deploy this are as follows:

- remove the existing records from the `rotation` and `rotation_cycle` tables - I didn't bother to script this, PhpMyAdmin to the rescue ;)
- import the new `rotations.json` with the usual import command
- run the `app:legality:decklists-rotation` command to move existing decklists to the newly created rotation id

I hope the last step is fast enough on the live server, I have no way to test its performance with ~46000 decks on my dev machine, unfortunately.


src/AppBundle/Command/LegalityDecklistsRotationCommand.php
```diff
@@ -42,6 +47,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                
+                $oldId = null;
+                if ($decklist->getRotation()) {
+                    $oldId = $decklist->getRotation()->getId();
+                }
+                
+                if ($confirm && $oldId !== $rotation->getId()) {
+                    $output->writeln("  updating decklist " . $decklist->getId());
+                    $decklist->setRotation($rotation);
+                    $this->entityManager->persist($decklist);
```
alsciende: $decklist is already persisted

src/AppBundle/Service/RotationService.php
```diff
@@ -46,9 +46,11 @@ public function isRotationCompatible(Decklist $decklist, Rotation $rotation)
         foreach ($decklist->getSlots() as $slot) {
             $cycles[$slot->getCard()->getPack()->getCycle()->getCode()] = 1;
         }
-
-        return count(array_diff(array_keys($cycles), array_map(function (Cycle $cycle) {
-            return $cycle->getCode();
-        }, $rotation->getCycles()->toArray()))) === 0;
+        
+        $allCycles = array_map(function(Cycle $cycle) { return $cycle->getCode(); }, $this->entityManager->getRepository(Cycle::class)->findBy([]));
```
alsciende: No need for that. We have `array_keys($cycles)` the list of cycles used by the decklist, and `$rotatedCycles` the list of cycles forbidden by the rotation. We can use `array_intersect($rotatedCycles, array_keys($cycles))` to find the forbidden cycles used by the decklist.

alsciende:
Actually I think using `src/AppBundle/Command/LegalityDecklistsRotationCommand.php` is a problem. That command does not seem to be valid anymore. In fact, the property `cycle.rotated` should not exist anymore, unless I'm mistaken. Where we're checking `cycle.rotated`, we should instead use `RotationService` to find out if the cycle is accepted in the latest rotation. Maybe `Rotation` should have an `active` property, like `Mwl`.

gereons:
It worked on my machine :) - I'll give this another go soon.

alsciende:
I'm sure it works :) It's more of a design problem.

Another problem is the performance. We must do our best not to use the ORM for such tasks.

Can you push your branch to this repo? I'll see if I can continue your work.

gereons:
I don't have write permission here, that would need Collaborator rights if I'm not mistaken. 
I've merged master into my brach, so if you checkout the branch from https://github.com/gereons/netrunnerdb you should be up-to-date.

gereons:
We're not going to need this anymore 😭

PR #241, opened 2018-09-12
allow moderators to delete comments on any decklist

bblum:
The 'About' page states that moderators should be able to delete comments as well as decklists. This pull request actually implements that feature, which would be useful on occasions when neither a decklist's author nor the site admin are promptly available to handle offensive posts.

I tested the client-side delete button stuff using element inspector just enough to find the server-side access check, but stopped short of actually building a local server instance, so if I missed any cases, please let me know.


alsciende (approved):

alsciende:
Thank you. Keep in mind that this will merely hide the comment, not delete it.

bblum:
Thanks. Yeah, I was gonna address that next, thought I'd start with a warm up.

PR #243, opened 2018-09-13
Request cardgamedb images from http not https

ejpeeZ:
Quick fix for https://github.com/Alsciende/netrunnerdb/issues/242

_Small disclaimer: I am a fledgeling web-developer, so I might not have seen the bigger picture here._

The `https` was added to prevent mixed content warnings, but now image requests to cardgamedb fail somehow. Changing it back to `http` works, but yeah, merge at your own discretion regarding the warnings.

### DIFF
--- a/src/AppBundle/Command/ImportImagesCommand.php
+++ b/src/AppBundle/Command/ImportImagesCommand.php
@@ -53,7 +53,7 @@ class ImportImagesCommand extends ContainerAwareCommand
             }
 
             $url = sprintf(
-                'https://www.cardgamedb.com/forums/uploads/an/med_ADN%d_%s.png',
+                'http://www.cardgamedb.com/forums/uploads/an/med_ADN%d_%s.png',
                 $ffgId,
                 $position
             );


maartenleenders:
Ugh I now realise this does not work if the imageUrls in the cards are not empty.

ejpeeZ:
Ugh I now realise this does not work if the imageUrls in the cards are not empty.

alsciende (approved):

gereons:
@Alsciende could you please run the image importer command on the live server to make this change effective? Plus, if you get around to do that, could you take a look at the two open PRs in the cards repo for the Errata and Magnum Opus cards? TIA!

ejpeeZ:
EDIT: it seems that the card images are back, so yay! :)

_Hmm but wouldn't running the importer command again do nothing now, because of line 42:_
```
$cards = $this->entityManager->getRepository(Card::class)->findBy(['imageUrl' => null]);
```

_Probably needs some_
```findBy(['imageUrl' => [ null, >>>code that finds strings that start with https://www.cardgamedb.com<<<< ] );```
But again, I can't test currently, and I am not familiar with this Doctrine thing.

_Or you could actually only find those strings that start with https://www.cardgamedb.com, because no new cards will be created by FFG, so there would be no `null` cards anyways_

gereons:
Thanks, @Alsciende ! Hi-rez scans for the Magnum Opus cards are here: https://drive.google.com/drive/folders/1HTIMOPQwjCRUVPTCiOr26vOys0SlVR-c

PR #254, opened 2018-11-26
Encoded SC19 glyph into font files and updated site to use it.

x3N1GM4x:
Figured out how icomoon.io was being used to generate the glyphs for the cycle icons (amongst other things) and added a new one for SC19:

image


x3N1GM4x:
`composer.json` and `composer.lock` shouldn't be in this PR and neither should `.gitignore` probably. I'm struggling to figure out the cleanest way to rectify this, so if anyone's able to help, please let me know and don't merge this PR until it's sorted!

NoahTheDuke:
Cleaning up your gitignore changes is best, to keep folks from having to make those same changes in their own: `web/card_image/` will work. You don't need to do the star, cuz it'll ignore the whole folder which is what we want.

To remove the other files, do `git rm --cached FILENAME` to remove them from the tree without removing them from your repo. I don't know the best solution to using the web server without adding them to composer, but honestly it might not be a bad idea to get everyone set up with it off the bat.

How's your experience of it been so far?

x3N1GM4x:
Thanks for that, how does that look now?

I've not had any problems with the php built in web server (asides from getting the dang thing to install in the first place) - it seems to do the job just fine.

x3N1GM4x:
Ok, hopefully we're good to go now?

NoahTheDuke:
Looks great!

NoahTheDuke (approved):
Looks great! Keep up the good work!

PR #255, opened 2018-11-30
Fixed SC19 for deckbuilding quantities

x3N1GM4x:
Appended sc19 to two hardcoded lists of "core sets", so it is handled correctly when users set a number of cores on the settings tab of the deckbuilder.

There are some other places (like the validation for Cache Refresh and Onesies) where we should also review core-validation, but this will do for now.

### DIFF
--- a/web/js/deck.js
+++ b/web/js/deck.js
@@ -10,7 +10,7 @@ function update_max_qty() {
     NRDB.data.cards.find().forEach(function(card) {
         var modifiedCard = get_mwl_modified_card(card);
         var max_qty = modifiedCard.deck_limit;
-        if(card.pack_code == 'core' || card.pack_code == 'core2') {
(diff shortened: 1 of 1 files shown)


web/js/deck.js
```diff
@@ -597,7 +597,7 @@ function handle_quantity_change(event) {
 function update_core_sets() {
 	CardDivs = [ null, {}, {}, {} ];
 	NRDB.data.cards.find({
-		pack_code : ['core', 'core2']
+		pack_code : ['core', 'core2','sc19']
```
binarymelon, line 600: Completely nitpicking, but maybe just add a space before 'sc19'.

x3N1GM4x: Yeah, I spotted this as well, thanks.
```suggestion
		pack_code : ['core', 'core2', 'sc19']
```

NoahTheDuke: Sorry for merging too fast! I'll add the fix.

PR #258, opened 2018-12-01
Deckbuilder fixes

NoahTheDuke:
* Moves "unicorn" check to function, generating for IDs now
* Updates `convert_to_rcs` to `convert_to_recent`, updates card list for System Core 2019
* Adds System Core 2019 to Onesies and Cache Refresh
* Cleans up indentation and trailing spaces


NoahTheDuke:
@x3N1GM4x Could you look at this to make sure I didn't break anything? I can split just the convert_to_recent change into a new PR if the other changes are too big.

web/js/nrdb.js
```diff
@@ -252,6 +251,12 @@ function find_identity() {
     Identity = NRDB.data.cards.find({ indeck: { '$gt': 0 }, type_code: 'identity' }).pop();
 }
 
+function unicorn(card) {
+    var mwlCard = get_mwl_modified_card(card);
+    var unicorn_emoji = '<span title="Restricted card" style="display:inline-block;width:1.5em;">🦄</span> ';
```
x3N1GM4x:
```suggestion
    var unicorn_emoji = ' <span title="Restricted card" style="display:inline-block;width:1.5em;">🦄</span> ';
```

web/js/nrdb.js
```diff
@@ -416,9 +422,7 @@ function update_deck(options) {
             additional_info = '(<span class="small icon icon-' + card.pack.cycle.code + '"></span> ' + card.position + ') ' + alert_number_of_sets + influence;
         }
 
-        var mwlCard = get_mwl_modified_card(card);
-        var unicorn = mwlCard.is_restricted ? '<span title="Restricted card" style="display:inline-block;width:1.5em;">🦄</span> ' : '';
-        var item = $('<div>' + card.indeck + 'x <a href="' + Routing.generate('cards_zoom', { card_code: card.code }) + '" class="card" data-toggle="modal" data-remote="false" data-target="#cardModal" data-index="' + card.code + '">' + card.title + '</a> ' + unicorn + additional_info + '</div>');
+        var item = $('<div>' + card.indeck + 'x <a href="' + Routing.generate('cards_zoom', { card_code: card.code }) + '" class="card" data-toggle="modal" data-remote="false" data-target="#cardModal" data-index="' + card.code + '">' + card.title + '</a> ' + unicorn(card) + additional_info + '</div>');
```
x3N1GM4x:
```suggestion
        var item = $('<div>' + card.indeck + 'x <a href="' + Routing.generate('cards_zoom', { card_code: card.code }) + '" class="card" data-toggle="modal" data-remote="false" data-target="#cardModal" data-index="' + card.code + '">' + card.title + '</a>' + unicorn(card) + additional_info + '</div>');
```

x3N1GM4x:
As discussed:

- Could we add a leading space when appending the unicorn emoji, otherwise it looks a bit squashed, especially against an ID. Have added some suggestions to the code which I think resolve this. Could maybe be even more elegant as I think what I've suggested will result in no spacing between the end of a card description and any influence pips when the card is not a unicorn, so might be worth revisiting for complete consistency.

- (Logged under #263) While cards are being correctly identified as rotated and the "Deck contains rotated cards" message is being correctly shown, clicking on the subsequent "click to update" link results in the label changing to "Replaced 0 card(s) with their post-rotation counterparts." and no changes being made to the offending card(s) in the deck. I have a feeling this was broken before this PR (like 95% sure), so maybe now would be a good time to knock that bug on the head too if we can? For example, adding the Revised Core Femme Fatale to a deck and then clicking the update link does nothing:
image
image

NoahTheDuke:
Rotation button: Definitely broken, probably broken beforehand as well. I'll pull that out and move it to a new PR so we can focus on it directly and get the unicorn fix out.

binarymelon (approved):

Issue #256, opened 2018-11-30
Restricted ID in deckbuilder not clear


x3N1GM4x:
If you build a deck with an Identity that is Restricted (for example, Mti on the latest MWL3.0) and then add another restricted card to that deck, you are correctly warned that you cannot have multiple restricted cards in the same deck.

However, it is not made entirely clear when one of those cards is an ID, which could lead to some confusion for users, especially if they're not familiar with the MWL (which is surely the target audience of this feature) or if they have multiple restricted cards in their deck as well as their ID being restricted.

Could we somehow mark IDs as being Restricted/unicorn to mirror what we do for all other cards in a deck?

x3N1GM4x:
So a couple of ideas how this could be done, firstly with an extra unicorn icon:

![image](https://user-images.githubusercontent.com/10621139/49309132-1f493100-f4d2-11e8-8a1a-b09f485a7249.png)
![image](https://user-images.githubusercontent.com/10621139/49309186-43a50d80-f4d2-11e8-945c-a8bca71d5b5b.png)

Or alternatively, perhaps by listing the specific cards which are restricted as part of the warning (or offering them in a click-through or mouseover tooltip of some kind)?

NoahTheDuke:
I think adding the unicorn to the ID name is the perfect way to handle this. It's already used and will be immediately obvious.

x3N1GM4x:
Fixed in #258.

PR #262, opened 2018-12-02
Added unicorn to ID title.

x3N1GM4x:
As discussed in #256, added unicorn icon to IDs:

image


x3N1GM4x:
Oops, this is already done (and more elegantly) in #258.

PR #265, opened 2018-12-03
Added Cycle Name after Datapack Name in tooltips.

### LINKED ISSUE #229: Feature request - show cycle of cards in data packs on hover tooltips and card pages
Hey! I'm still pretty new to Netrunner, so I'm often not sure which cycle a card belongs to. For instance, take this hover tooltip for Reaver:

I'm not sure which cycle Reaver is in (off the top of my head, anyway), and to check I have to pull up the [search page](https://netrunnerdb.com/en/search), since it's got its nice list of sets: 

It'd be a bit easier for folks newer to Netrunner if they could see the cycle a data pack is in, like so:

> ●●●● Apex - Martial Law (Flashpoint)

Is this something you'd be open to incorporating / receiving a PR on?

x3N1GM4x:
Resolves #229:

image
image
image
image

### DIFF
--- a/web/js/nrdb.tip.js
+++ b/web/js/nrdb.tip.js
@@ -22,8 +22,16 @@
                         /.*\/card\/(\d\d\d\d\d).*/,
                         "$1"));
         var card = NRDB.data.cards.findById(code);
+        
         if(!card)
             return;
+        var cycle = NRDB.data.cycles.findById(card.pack.cycle_code);
+        var cycle_name = '';
+        if(cycle.size === 1) {
+            cycle_name = ''
+        } else {
+            cycle_name = ' (' + card.pack.cycle.name + ')'
+        }
         var type = '<p class="card-info">' + NRDB.format.type(card) + '</p>';
         var influence = '';
         for(var i = 0; i < card.faction_cost; i++)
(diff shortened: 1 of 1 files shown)


web/js/nrdb.tip.js
```diff
@@ -22,8 +22,16 @@
                         /.*\/card\/(\d\d\d\d\d).*/,
                         "$1"));
         var card = NRDB.data.cards.findById(code);
+
```
NoahTheDuke: Whitespace: 🙅‍♂️

web/js/nrdb.tip.js
```diff
@@ -22,8 +22,16 @@
                         /.*\/card\/(\d\d\d\d\d).*/,
                         "$1"));
         var card = NRDB.data.cards.findById(code);
+        
         if(!card)
             return;
+        var cycle = NRDB.data.cycles.findById(card.pack.cycle_code);
+        var cycle_name = '';
+        if(cycle.size === 1) {
```
NoahTheDuke: Try `if(cycle.size !== 1) {` and then the rest. No need to reassign when you already assigned it blank above.

web/js/nrdb.tip.js
```diff
@@ -22,8 +22,16 @@
                         /.*\/card\/(\d\d\d\d\d).*/,
                         "$1"));
         var card = NRDB.data.cards.findById(code);
+        
         if(!card)
             return;
+        var cycle = NRDB.data.cycles.findById(card.pack.cycle_code);
```
binarymelon: Since it appears the cycle is already defined on the card do we need to do this lookup?

x3N1GM4x:
Ok, I think this is significantly cleaner/better, thanks for the tips!

web/js/nrdb.tip.js
x3N1GM4x: Yes, good point! I got sloppy. 😄

Issue #263, opened 2018-12-03
"click to update" on "Deck contains rotated cards" message doesn't work


x3N1GM4x:
Identified while checking #258.

If you start building a deck and add a rotated card to it, you are warned and offered a chance to update such cards:
![image](https://user-images.githubusercontent.com/10621139/49385677-de415e80-f715-11e8-83e1-248a87178408.png)
However, clicking on this link doesn't actually do anything:
![image](https://user-images.githubusercontent.com/10621139/49385690-e6999980-f715-11e8-9a57-ec8ab1c4492b.png)
Needs investigating and resolving so this feature actually works.

PR #267, opened 2018-12-03
Update Convert Deck to Recent Cards

NoahTheDuke:
Changes from #258 moved to a new PR to test more. Probably fine, who knows?


x3N1GM4x:
I pulled this onto a fresh local branch and it seems to be working fine now. Will do a few more specific tests before merging, but reckon this will be fine...
image

x3N1GM4x:
Seems to only work once a deck has been saved (for me at least), but this is still an improvement on it not working at all, so seems good to me! 😄

x3N1GM4x:
This fixes #263 as well.

PR #269, opened 2018-12-08
Allow publishing decks even when they contain unreleased cards

gereons:
Currently, NRDB does not allow publishing decks that contain unrelased cards. This to me feels odd and unnecessary, particulary now where the cards from SC19 are technically not released although the huge majority of the players already own them.

This PR removes the restriction.

### DIFF
--- a/src/AppBundle/Controller/SocialController.php
+++ b/src/AppBundle/Controller/SocialController.php
@@ -48,15 +48,6 @@ class SocialController extends Controller
             throw $this->createAccessDeniedException();
         }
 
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
This is awesome, thank you!

x3N1GM4x (approved):
Yep, this is great, thanks for doing this!

NoahTheDuke:
[Allow publishing decks that have unreleased cards](https://trello.com/c/Cm5TStN3/55-allow-publishing-decks-that-have-unreleased-cards)

PR #281, opened 2018-12-25
Improved speed of decklist search with cards tenfold

NoahTheDuke:
Previously: `23 rows in set (11.16 sec)`
Now: `23 rows in set (0.09 sec)`


NoahTheDuke:
Put all this technology into the other one, #277 .

Issue #252, opened 2018-11-13
Cannot Publish Card Review


Agent-Chezni:
Hello and thanks for all you do!

I got a bit inspired and wrote an 8,000 character review on Awakening Center...

I have it saved locally, but when I go to post it, I get the error "An error occurred while posting your review (OK). Reload the page and try again."

Is it being rejected b/c of the character count, or is there something else?

I've attached a .txt file of the text in case it helps.
Thanks!
--Chez

[Awakening Center.txt](https://github.com/Alsciende/netrunnerdb/files/2574798/Awakening.Center.txt)

NoahTheDuke:
Holy shit lol, that’s amazing. Let me check, I do believe there is a character limit but I don’t remember off the top of my head.

Saintis:
I believe a limit was put in when someone posted a decklist with the full text of Moby Dick in the description...

x3N1GM4x:
I tried to reproduce the issue on my local instance, but it posted the review without issue. I'm reluctant to test in production right now as I don't think there's a front-end mechanism for removing a review (at least not for non-admins like me).

It looks like the full error is written to the console, lines 231-234 of `zoom.js`:

```
error: function (jqXHR, textStatus, errorThrown) {
    console.log('[' + moment().format('YYYY-MM-DD HH:mm:ss') + '] Error on ' + this.url, textStatus, errorThrown);
    form.replaceWith('<div class="alert alert-danger" role="alert">An error occured while posting your review (' + jqXHR.statusText + '). Reload the page and try again.</div>');
}
```

@Agent-Chezni - can you confirm that you still get this issue if you try to post the review now and if so, can you check the console within your browser for any message of the format `[YYYY-MM-DD HH:mm:ss] Error on {something}` ?

ghost:
I got this too and wrote an articel about 3x times. Had to login explicitly to avoid it from happening.

x3N1GM4x:
Closing as should be resolved by #285.
@Agent-Chezni  - let us know if you're still having this problem.

Issue #250, opened 2018-10-28
Ika Review


Havvy86:
It appears you cannot post a review for Ika card. My username is Havvy and I believe I have the correct rights to post the review?

x3N1GM4x:
Have you been able to post reviews on other cards since you noticed this? Do you get any specific error message or behaviour you can describe (or event better, capture in screenshots) when you attempt to post the review?

There is also one review on the live site for this card, so it's not a universal inability to post reviews for this card.

I'm not able to reproduce this issue on my local instance, so we're going to need to try and gather some more information to be able to help.

Havvy86:
Hi there!

(By any chance, are you Ian Witt? I’m in the tournament group!).

I’ve tried this again today and got the following error:

Reviews
An error occured while posting your review (OK). Reload the page and try again.
I noticed there are reviews on this card now, so I suspect it might be just me?

Thanks

Ian

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

x3N1GM4x:
Yes, that's me, hi!

It certainly sounds like something local to your machine or account. Perhaps you could take a look at your web browser's console when you get that error to see if any further detail is being recorded there?

Details on how to do this if you're not aware will vary depending on your browser, there's a summary for the major browsers [here](https://documentation.concrete5.org/tutorials/how-open-browser-console-view-errors). Alternatively, just google "[browser_name] view console" or similar.

ghost:
See #252 as I assume it's the same and as there I think this is linked to #247

Havvy86:
That looks identical with the error message.

Word count is 243 though for my review!

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

x3N1GM4x:
Closing as should be resolved by #285.
@Havvy86 - let us know if you're still having this problem.

Issue #247, opened 2018-10-05
Need to re-login when deleting deck


tf42src:
Is this a feature? I am logged in, I can browse my decks etc. but deleting a deck prompts me with the login screen again.

x3N1GM4x:
It's not intentional as far as I'm aware, and I'm also unable to reproduce it either in Production or on my local instance. Have you always been re-prompted for credentials when deleting a deck, or has this only started happening recently?

Do you still have the problem if you log out of NRDB, log back in and immediately attempt to delete a deck once logged in? I have experienced my session timing out while being sat on deckbuilder or some other screen for too long.

tf42src:
Started a few months ago I think. Definitely shouldn't have hit session timeout as I (sometimes) just had logged in before.

gereons:
Can confirm, this has happened to me a few times as well.

ghost:
Same for me, it's annoying!
I think it has to do with the same issue like writing reviews for cards #252 , but this is only a feeling!
And I have the same using ABR, whenever I want to claim I have to relogin.

gereons:
Just happened to me again, which prompted some digging. AFAICT, this is not a bug but rather an intentional feature. 

Symphony apparently distingushes between `IS_AUTHENTICATED_FULLY` and `IS_AUTHENTICATED_REMEMBERED` login states - you're "fully" authenticated when you've just entered your credentials during this session, and "remembered" when you're logged in because you have the correct cookies. Now, the methods responsible for deleting decks (`BuilderController::deleteAction` and `BuilderController::deleteListAction`) are marked `@IsGranted("IS_AUTHENTICATED_FULLY")`, so you're forced to re-login.

NoahTheDuke:
That’s super interesting! Sounds like a protection against cookie stealing or hijacking.

gereons:
Most likey that's the intention. 

The UX is still crappy as hell IMO; if a site does something like this it should a) explain this to me, and (more importantly) b) don't make me f****ing re-do all my input, in this case re-select all the decks I'd like to remove.

ghost:
I would not call it an enhancement! When looking into the write card review, the site terminates with an error and all your input is lost! I wrote an article 3x!
So I think something should happen around it, as a noob has no chance to see where it's coming from and might be frustrated by the site!
So a chance to login would be very very welcome, but losing content people are providing is not something where we need an enhancement.

gereons:
[...]

PR #285, opened 2018-12-27
change authentication requirement from FULLY to REMEMBERED

gereons:
This should fix the bad UX shown in #247, #250 and #252.

I didn't change the annotation for `saveDeckAction` in `PrivateApi20Controller.php` because a) I have no idea how to test private API access locally, and b) I have never experienced a problem using this API from Net Deck.


NoahTheDuke:
This doesn't require any migrations, right? Just pull the changes and it'll automatically reload?

gereons:
I think so, yes. Maybe flushing the symphony cache is needed, I can't really tell from googling, but nothing more than that.

NoahTheDuke:
I'll make sure to do that when I pull this in.

PR #292, opened 2019-01-03
Navbar, Syntax, and Card changes

NoahTheDuke:
This change does a number things:
1) Removes the Rules pages because it is woefully out of date and not in line with what NRDB is for
2) Splits the previous Cards search page into two parts, a Sets page that displays information about each release and an Advanced Search page that is the meat of the old card search page
3) Adds a Syntax page that describes the search filter syntax in more detail than in a hover or little button
4) Revamps the header/navbar to display these changes

Navbar:
screen shot 2019-01-02 at 7 53 18 pm
Sets page:
screen shot 2019-01-02 at 8 05 42 pm
Advanced Search:
screen shot 2019-01-02 at 8 12 08 pm
Syntax description
screen shot 2019-01-02 at 8 15 10 pm


x3N1GM4x:
Added an update to slightly indent Packs on the new Sets listing page:

image

Sorry for the superfluous merge commits - I had some trouble figuring out how to push to this feature branch on my upstream!

app/Resources/views/Search/display-set.html.twig
```diff
@@ -0,0 +1,62 @@
...
+
+{% block body %}
+<div class="container">
+   <div id="list">
+      <table class="rwd-table table table-striped table-condensed">
+         <thead>
+            <tr>
+               <th>Name</th>
+               <th>Cards</th>
+               <th>Date</th>
```
x3N1GM4x: Is it worth being more specific here? Very nitpicky I know!

```suggestion
                <th>Release Date</th>
```

PR #294, opened 2019-01-03
Fix issue with PAD Factory influence with sc19 PAD Campaign

### LINKED ISSUE #293: PAD Campaign from System Core Set does not satisfy alliance requirements for PAD Factory
Repeat of an old issue when revised core set was introduced:

When using System Core Set PAD Campaigns, the alliance requirement is not considered fulfilled for the purposes of PAD Factory.

x3N1GM4x:
Fixes #293.


x3N1GM4x:
Again not entirely clear on how I've ended up with some extra unexpected commits on this PR, but hopefully as they're already in `master` it doesn't matter.

NoahTheDuke:
I'll squash it!

PR #303, opened 2019-01-20
Add a field for patreon pledges

Absotively:
An attempt to handle Patreon pledges while leaving the current donation info untouched.

Requires a schema update. Output of `php bin/console doctrine:schema:update --dump-sql` is:

>     ALTER TABLE user ADD patreon_pledge_cents INT DEFAULT 0 NOT NULL;

So if we do use this, that needs to be run on production before the code there is updated, I guess? I'm still not 100% clear on how Symfony's schema management works.

### DIFF
--- a/src/AppBundle/Command/DonatorCommand.php
+++ b/src/AppBundle/Command/DonatorCommand.php
@@ -60,7 +60,7 @@ class DonatorCommand extends ContainerAwareCommand
                 if ($type == 'paypal') {
                     $user->setDonation($amount + $user->getDonation());
                 } else {
-                    $user->setPledgeCents($amount * 100);
+                    $user->setPatreonPledgeCents($amount * 100);
                 }
                 $this->entityManager->flush();
                 $output->writeln(date('c') . " " . "Success");
@@ -68,7 +68,7 @@ class DonatorCommand extends ContainerAwareCommand
                 $output->writeln(date('c') . " " . "Cannot find user [$email]");
             }
         } else {
-            $output->writeln(date('c') . " " . "Invalid donation type");
+            $output->writeln(date('c') . " " . "Invalid donation type [$type]: expected 'patreon' or 'paypal'");
         }
     }
 }
--- a/src/AppBundle/Command/DonatorCommand.php
+++ b/src/AppBundle/Command/DonatorCommand.php
@@ -31,16 +31,22 @@ class DonatorCommand extends ContainerAwareCommand
                 'Email address or username of user'
             )
             ->addArgument(
-                'donation',
+                'type',
                 InputArgument::REQUIRED,
-                'Amount of donation'
+                '"paypal" or "patreon"'
+            )
+            ->addArgument(
+                'amount',
+                InputArgument::REQUIRED,
+                'Amount of donation in US dollars'
             );
     }
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $email = $input->getArgument('email');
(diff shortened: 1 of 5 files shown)


src/AppBundle/Entity/User.php
```diff
@@ -298,6 +301,33 @@ public function setDonation(int $donation)
         return $this;
     }
     
+    /**
+     * @return integer
+     */
+    public function getPledgeCents()
```
plural: any reason to have a slight difference between the field and method names here?  Are we anticipating other pledge types or will this be patreon?

src/AppBundle/Command/DonatorCommand.php
```diff
@@ -49,12 +55,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                    $user->setPledgeCents($amount * 100);
+                }
+                $this->entityManager->flush();
+                $output->writeln(date('c') . " " . "Success");
+            } else {
+                $output->writeln(date('c') . " " . "Cannot find user [$email]");
+            }
         } else {
-            $output->writeln(date('c') . " " . "Cannot find user [$email]");
+            $output->writeln(date('c') . " " . "Invalid donation type");
```
plural: add the donation type supplied to the error message?

src/AppBundle/Service/PersonalizationHelper.php
```diff
@@ -47,7 +47,7 @@ public function defaultBlock(User $user)
             'introductions'      => $user->getIntroductions(),
             'faction'            => $user->getFaction(),
             'autoload_images'    => $user->getAutoloadImages(),
-            'donation'           => $user->getDonation(),
+            'is_supporter'       => $user->isSupporter(),
```
plural, line 50: i couldn't quite tell if the donation field from this is used anywhere or not.  I don't think so.  Any reason to have this if it isn't used?

plural (commented):
hi! New webdev team member jumping on here with some questions and comments.  I'm still navigating some of the PHP symfony stuff, so please forgive any silly questions.

src/AppBundle/Entity/User.php
Absotively: Mostly I was just making it shorter. I don't think we're planning to have any other pledge types, but adding Patreon to the method name probably isn't a bad idea. I'll do that.

src/AppBundle/Service/PersonalizationHelper.php
Absotively: It isn't used, and it actually isn't there - that's how GitHub shows deleted lines in diffs.

src/AppBundle/Entity/User.php
Absotively: Fixed.

src/AppBundle/Command/DonatorCommand.php
Absotively: I don't think this is likely to be an issue, since it's for internal use and the option is right there in the command line above the error message, but it also wasn't hard to add, so I did.

src/AppBundle/Service/PersonalizationHelper.php
plural: yeah, brain freeze.  Sorry!

plural (commented):
thanks for rolling with the drive by comments!

plural (approved):

NoahTheDuke:
This looks great. Thank you so much for doing this!

PR #322, opened 2019-02-25
Fix the sort function for the deck building page.

### LINKED ISSUE #306: Sort on Build tab of Deckbuilder page doesn't work
Clicking the column headers doesn't seem to do anything any more:

image

Tested on all columns, both Chrome and Firefox, two different users.

plural:
select_only_latest_cards should not do filtering *and* sorting.

Fixes #306

### DIFF
--- a/web/js/deck.js
+++ b/web/js/deck.js
@@ -100,13 +100,10 @@ Promise.all([NRDB.data.promise, NRDB.settings.promise]).then(function() {
 
 	function findMatches(q, cb) {
 		if(q.match(/^\w:/)) return;
-		var matchingPacks = NRDB.data.cards.find({pack_code: Filters.pack_code || []});
-		var latestCards = select_only_latest_cards(matchingPacks);
 		var regexp = new RegExp(q, 'i');
-		var matchingCards = _.filter(latestCards, function (card) {
-			return regexp.test(_.deburr(card.title).toLowerCase().trim());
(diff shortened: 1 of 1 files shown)


x3N1GM4x:
Looks good, nice one!

NoahTheDuke:
Where is `normalized_title` coming from? It’s not defined for cards in the card data.

And without the `_.deburr` line, you lose the fix for searching for cards like Déja Vu and Sifr by name in the deck search.

x3N1GM4x:
It's almost like I didn't do any regression testing before I merged, just checked it fixed the bug. 😛

NoahTheDuke:
It’s all good, NRDB is arcane and byzantine.

plural:
@NoahTheDuke : i'm seeing normalized_title populated in the results of NRDB.data.cards.find() calls and the type-ahead in the search box on the building page is still working.  I'm definitely not all that familiar with the normalized_title (except why it exists).  Are there other troubles here because of this patch or is all well?

NoahTheDuke:
We already discussed this on slack, but to keep history public, `normalized_title` was removed in PR #310, as I had trouble invalidating users' cache, making it impossible to introduce a new field to the cards data.

x3N1GM4x:
> NRDB is arcane and byzantine.

And now I learned a new word too. 😝

PR #323, opened 2019-02-26
Remove normalized_title stuff that i snuck back in.

plural:
Sorry for the bad pull request. I was starting from an old HEAD CL.


NoahTheDuke:
This is great. Would you mind adding the sorting back in here? Not to `select_only_latest_cards`, but in `findMatches`, at the end. It’s needed for the typeahead drop down.

plural:
ok, added sorting back.

PR #327, opened 2019-02-26
Update select_only_latest_cards to preserve card order

NoahTheDuke:
Updates other uses of `select_only_latest_cards` to match @plural's fixes.

### DIFF
--- a/src/AppBundle/Service/CardsData.php
+++ b/src/AppBundle/Service/CardsData.php
@@ -952,11 +952,9 @@ class CardsData
             }
         }
 
-        $cardFilter = function ($card, $key) use ($latestCardsByTitle) {
+        return array_values(array_filter($matchingCards, function ($card) use ($latestCardsByTitle) {
             return $card->getCode() == $latestCardsByTitle[$card->getTitle()]->getCode();
-        };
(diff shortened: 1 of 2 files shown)


src/AppBundle/Service/CardsData.php
```diff
@@ -951,7 +951,12 @@ public function select_only_latest_cards(array $matchingCards)
                 $latestCardsByTitle[$title] = $card;
             }
         }
-        return array_values($latestCardsByTitle);
+
+        $cardFilter = function ($card, $key) use ($latestCardsByTitle) {
```
plural: i'm not familiar with use in this context.  What is that doing?

NoahTheDuke: (I've moved this function into the `array_filter` call directly, to be clear.)

Functions in php define a new empty scope and can't access external variables, so like Python 3's `nonlocal`, `use ($variable)` brings `$variable` into the function's scope without passing it through a parameter. It allows for writing closures over local variables.

plural: Ah, thanks for the clarification!

PR #329, opened 2019-02-26
Refactor Search Options

NoahTheDuke:
Currently, the two search option lists have problems. "Sort by Set" is ambiguous: is it by release date or by set name? And View Rulings is broken. Additionally, the searchoptions twig is ugly as sin and hard to work with.

This refactors the entire file, making it more consistent and easier to manage. I changed "Sort by Set" to unambiguously be "Sort by Set Name" and added a "Sort by Release Date" option. I also removed the "View Rulings" option as it's out of scope for this PR to fix and it's bad to leave broken options available for use and changed "Simple List" to be "Names only", which is more accurate and clear.

I also made both lists somewhat consistent in their language: "Sort by {{blank}}" and "View as {{blank}}".

### Before:
Sort: 
screen shot 2019-03-05 at 10 44 42 am
View:
screen shot 2019-03-05 at 10 44 46 am

### After
Sort:
screen shot 2019-03-05 at 10 45 02 am
View:
image

### DIFF
--- a/app/Resources/views/Search/searchoptions.html.twig
+++ b/app/Resources/views/Search/searchoptions.html.twig
@@ -1,15 +1,6 @@
-{% set sort_options = {
-	'name': 'Name',
-	'set': 'Set Name',
-	'release-date': 'Release Date',
-	'faction': 'Faction',
-	'type': 'Type',
-	'cost': 'Cost',
-	'strength': 'Strength',
-} %}
 <div class="col-sm-4" style="margin-bottom:.5em">
 	<select name="sort" class="form-control">
-		{% for key, value in sort_options|cast_to_array %}
+		{% for key, value in sort_options %}
 			<option value={{key}} {% if sort == key %}selected{% endif %}>
 			Sort by {{value}}
 			</option>
@@ -17,16 +8,9 @@
 	</select>
 </div>
 
-{% set view_options = {
-	'list': 'a Checklist',
-	'text': 'a Spoiler',
-	'full': 'Full Cards',
-	'images': 'Images only',
-	'short': 'Names only',
-} %}
 <div class="col-sm-4" style="margin-bottom:.5em">
 	<select name="view" class="form-control">
-		{% for key, value in view_options|cast_to_array %}
+		{% for key, value in view_options %}
 			<option value={{key}} {% if view == key %}selected{% endif %}>
 			View as {{value}}
 			</option>
--- a/app/Resources/views/Search/searchoptions.html.twig
+++ b/app/Resources/views/Search/searchoptions.html.twig
@@ -1,21 +1,39 @@
+{% set sort_options = {
+	'name': 'Name',
+	'set': 'Set Name',
+	'release-date': 'Release Date',
+	'faction': 'Faction',
+	'type': 'Type',
+	'cost': 'Cost',
+	'strength': 'Strength',
+} %}
 <div class="col-sm-4" style="margin-bottom:.5em">
-<select name="sort" class="form-control">
-	<option value="name"    {% if sort|default('name') == "name"    %}selected{% endif %}>Sort by Name</option>
-	<option value="set"     {% if sort|default('name') == "set"     %}selected{% endif %}>Sort by Set</option>
-	<option value="faction" {% if sort|default('name') == "faction" %}selected{% endif %}>Sort by Faction</option>
-	<option value="type"    {% if sort|default('name') == "type"    %}selected{% endif %}>Sort by Type</option>
-	<option value="cost"    {% if sort|default('name') == "cost"    %}selected{% endif %}>Sort by Cost</option>
-	<option value="strength"{% if sort|default('name') == "strength"%}selected{% endif %}>Sort by Strength</option>
-</select>
+	<select name="sort" class="form-control">
+		{% for key, value in sort_options|cast_to_array %}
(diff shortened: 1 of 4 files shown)


plural:
I have always viewed set as basically "release order".  Maybe both of them work and should be offered under different names?

NoahTheDuke:
I think that’s the best option. I’ll make that change.

app/Resources/views/Search/searchoptions.html.twig
```diff
@@ -1,21 +1,39 @@
+{% set sort_options = {
```
binarymelon: Would it make sense to pull the sort_options and view_options up to the Search controller?

binarymelon: Also is the filter needed?  Can't we just define them as an array of key-value pairs?

NoahTheDuke: If you have a code sample, I’ll try it out!

binarymelon: I don't have my nrdb environment up on my work laptop and my Symfony experience is practically non-existent, but I'm thinking maybe something like.
```diff
Index: src/AppBundle/Twig/AppExtension.php
===================================================================
--- src/AppBundle/Twig/AppExtension.php	(date 1551800862000)
+++ src/AppBundle/Twig/AppExtension.php	(date 1551800862000)
@@ -1,21 +0,0 @@
-<?php
-namespace AppBundle\Twig;
-
-use Twig\Extension\AbstractExtension;
-use Twig\TwigFilter;
-
-class AppExtension extends AbstractExtension
-{
-    public function getFilters()
-    {
-        return [
-            new TwigFilter('cast_to_array', [$this, 'objectFilter']),
-        ];
-    }
-
-    public function objectFilter($stdClassObject) {
-        $response = (array)$stdClassObject;
-
-        return $response;
-    }
-}
Index: app/Resources/views/Search/searchoptions.html.twig
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- app/Resources/views/Search/searchoptions.html.twig	(date 1551800862000)
+++ app/Resources/views/Search/searchoptions.html.twig	(date 1551807710893)
@@ -1,15 +1,6 @@
-{% set sort_options = {
-	'name': 'Name',
-	'set': 'Set Name',
-	'release-date': 'Release Date',
-	'faction': 'Faction',
-	'type': 'Type',
-	'cost': 'Cost',
-	'strength': 'Strength',
-} %}
 <div class="col-sm-4" style="margin-bottom:.5em">
 	<select name="sort" class="form-control">
-		{% for key, value in sort_options|cast_to_array %}
+		{% for key, value in sort_options %}
 			<option value={{key}} {% if sort == key %}selected{% endif %}>
 			Sort by {{value}}
 			</option>
@@ -17,16 +8,9 @@
 	</select>
 </div>
 
-{% set view_options = {
-	'list': 'a Checklist',
-	'text': 'a Spoiler',
-	'full': 'Full Cards',
-	'images': 'Images only',
-	'short': 'Names only',
-} %}
 <div class="col-sm-4" style="margin-bottom:.5em">
 	<select name="view" class="form-control">
-		{% for key, value in view_options|cast_to_array %}
+		{% for key, value in view_options %}
 			<option value={{key}} {% if view == key %}selected{% endif %}>
 			View as {{value}}
 			</option>
Index: src/AppBundle/Controller/SearchController.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/AppBundle/Controller/SearchController.php	(date 1551800862000)
+++ src/AppBundle/Controller/SearchController.php	(date 1551807394488)
@@ -14,6 +14,24 @@
 
 class SearchController extends Controller
 {
+    const SORT_OPTIONS = array(
+        'name' => 'Name',
+	    'set' => 'Set Name',
+	    'release-date' => 'Release Date',
+        'faction' => 'Faction',
+        'type' => 'Type',
+        'cost' => 'Cost',
+        'strength' => 'Strength'
+    );
+    
+    const VIEW_OPTIONS = array(
+        'list' => 'a Checklist',
+        'text' => 'a Spoiler',
+        'full' => 'Full Cards',
+        'images' => 'Images only',
+        'short' => 'Names only'
+    );
+    
     /**
      * @param EntityManagerInterface $entityManager
      * @param CardsData              $cardsData
@@ -84,6 +102,8 @@
             "types"           => $types,
             "keywords"        => $keywords,
             "illustrators"    => $illustrators,
+            "view_options"    => self::VIEW_OPTIONS,
+            "sort_options"    => self::SORT_OPTIONS
         ], $response);
     }
 
@@ -471,6 +491,8 @@
             "q"    => $q,
             "view" => $view,
             "sort" => $sort,
+            "view_options"    => self::VIEW_OPTIONS,
+            "sort_options"    => self::SORT_OPTIONS
         ]);
 
         if (empty($title)) {
```

NoahTheDuke: Oh, I see what you mean! So the filter was needed because by default, Twig can't iterate over objects, but I forgot that php they're arrays. Thanks! I've updated it to match your suggestion.

NoahTheDuke: And good call with advanced_search. I missed that usage.

PR #330, opened 2019-02-27
Firefox: Show card thumbnails when hovering over a card

### LINKED ISSUE #161: Card previews do not contain image thumbnails in Firefox 56.0.1
Hovering over a card link in Firefox 56.0.1 causes the card preview to appear. However, the preview does not contain the art thumbnail in the upper-right corner--neither the image itself nor its hexagonal frame.

---

netrunnerdb - firefox card preview
-*Card preview in Firefox 56.0.1*

---

When viewing in other browsers (Edge 38.14393.1066.0, Chrome 61.0.3163.100), the thumbnail and the hexagon appear as expected:

---

netrunnerdb - edge card preview
-*Card preview in Edge 38.14393.1066.0*

---

netrunnerdb - chrome card preview
-*Card preview in Chrome 61.0.3163.100*

---

I have tested this in Firefox 56.0.1 Safe Mode, with all add-ons disabled. The behavior is the same.

Is this a bug? Or is Firefox just not capable of displaying the thumbnails for some reason?

gereons:
As far as I can tell, there is no longer any reason to not display these nice thumbnails in Firefox.

fixes #161

### DIFF
--- a/web/js/nrdb.tip.js
+++ b/web/js/nrdb.tip.js
@@ -31,8 +31,7 @@
         if(card.strength != null)
             type += '<p>Strength <b>' + card.strength + '</b></p>';
         var image_svg = '';
-        if($('#nrdb_svg_hex').length && typeof InstallTrigger === 'undefined') {
-            // no hexagon for Firefox, bad boy!
+        if($('#nrdb_svg_hex').length) {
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Works like a dream!

PR #333, opened 2019-03-05
Add Patreon supports to Donator page

NoahTheDuke:
Folks who donate through Patreon aren't getting listed on the Donators page. This fixes that, showing them in the list based on their combined donations across both platforms.

### DIFF
--- a/src/AppBundle/Controller/SocialController.php
+++ b/src/AppBundle/Controller/SocialController.php
@@ -1144,7 +1144,7 @@ class SocialController extends Controller
 
         $dbh = $entityManager->getConnection();
 
-        $users = $dbh->executeQuery("SELECT * FROM user WHERE donation > 0 OR patreon_pledge_cents > 0 ORDER BY (donation + patreon_pledge_cents) DESC, username", [])->fetchAll(\PDO::FETCH_ASSOC);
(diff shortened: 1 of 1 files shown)


src/AppBundle/Controller/SocialController.php
```diff
@@ -1144,7 +1144,7 @@ public function donatorsAction(EntityManagerInterface $entityManager)
 
         $dbh = $entityManager->getConnection();
 
-        $users = $dbh->executeQuery("SELECT * FROM user WHERE donation>0 ORDER BY donation DESC, username", [])->fetchAll(\PDO::FETCH_ASSOC);
+        $users = $dbh->executeQuery("SELECT * FROM user WHERE donation > 0 OR patreon_pledge_cents > 0 ORDER BY (donation + patreon_pledge_cents) DESC, username", [])->fetchAll(\PDO::FETCH_ASSOC);
```
Absotively: Should order by `donation + 100*patreon_pledge_cents`, because `donation` is in dollars.

NoahTheDuke: Aw heck, good call! Thank you!

NoahTheDuke:
```suggestion
        $users = $dbh->executeQuery("SELECT * FROM user WHERE donation > 0 OR patreon_pledge_cents > 0 ORDER BY (donation + (patreon_pledge_cents * 100)) DESC, username", [])->fetchAll(\PDO::FETCH_ASSOC);
```

PR #345, opened 2019-03-26
Add Patreon button to nav bar

x3N1GM4x:
My first attempt at this - I am definitely no CSS/bootstrap expert, so I fully expect there to be a better way to do this, which I'd love someone to point out to me.

It took me a while to figure out where some of the CSS for the nav bar was coming from (twitter bootstrap) and how I could override it in our own css.

If someone could take a look and then show me how to achieve the same thing in a better way, that would be great!


x3N1GM4x:
[P2: Add Patreon badge to site](https://trello.com/c/Xecha9Nw/66-p2-add-patreon-badge-to-site)

NoahTheDuke:
The css looks fine, but I worry that placing it at the top (valuable real estate in an already cramped header) will feel grabby or overboard. I think I'd prefer to have it listed in the footer along with another image-link to donate with Paypal. There we can make them as big and flashy as we want, while not detracting overall from the usability of the site.

x3N1GM4x:
Sure, I had the same thoughts, but thought that placement consistency between nisei.net and NRDB might be worth it. Let's discuss with the team before making a decision.

plural:
i added a patreon button to the bottom section in a recent PR (#457) so i'm closing this out.

Issue #95, opened 2017-03-22
deck/view/ does not show MWL influence stars and lowered influence


dranimrev:
When editing deck, stars are visible and influence is correctly lowered and displayed. 
When going to view mode on finished deck, stars are not visible and influence is not lowered accordingly to MWL. 
Legality was set to: NAPD MWL 1.1 (active)

![los](https://cloud.githubusercontent.com/assets/19939974/24200644/07481612-0f0e-11e7-9821-18fad5cb371c.jpg)

alsciende:
Duplicate of https://github.com/Alsciende/netrunnerdb/issues/73

plural:
ok, this has been fixed and deployed!

PR #356, opened 2019-05-27
Use the correct variable for the MWL on the deck view page.

plural:
Since the MWL code was undefined, the status indicators aren't present on the deck view page.  This includes things like the unicorn icon and the older MWL stars.

This fixes issue #95.

Before:
Screenshot 2019-05-27 at 10 52 08 AM

After:
Screenshot 2019-05-27 at 10 56 02 AM

### DIFF
--- a/web/js/deckview.js
+++ b/web/js/deckview.js
@@ -13,7 +13,7 @@ $(document).on('data.app', function() {
 		});
 	});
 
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Can you make it two separate lines? Having it be separated by a comma makes it really hard to read.

plural:
Noah, i tidied that up a bit and removed an extra temporary variable.

It is probably worth doing a similar cleanup in other places in a different changelist.

Issue #50, opened 2016-09-13
DotW decklist on the frontpage missing MWL influence stars


zaroth:
Example here: http://imgur.com/RbNF6vE , breaking news should have stars in front of them, they don't. The bug persists only in the frontpage view of the deck, everything's fine after you go to the deck page: http://imgur.com/ZnFztI9

plural:
ooh, this is still happening!

The current (as of  2019-05-12) DotW is missing the unicorn on Global Food Initiative.
![Screenshot 2019-05-12 at 4 03 18 PM](https://user-images.githubusercontent.com/396562/57587746-a2ce6b80-74cf-11e9-9e24-ecd38f553fac.png)

vs the deck page:
![Screenshot 2019-05-12 at 4 03 31 PM](https://user-images.githubusercontent.com/396562/57587747-a82bb600-74cf-11e9-9dc6-94e03014014c.png)

plural:
The fix for this has been deployed to prod.  I re-ran the highlight command so the current deck shows the MWL icons properly.

PR #357, opened 2019-05-27
Fix the MWL usage on the deck of the week view on the homepage.

plural:
The MWL code was not part of the custom JSON that the highlight
command uses.  Additionally, the javascript would not try to
initialize an MWL instance even if it was there.

This has all been fixed and a bit of tidying up on the index page
and the highlight command was done along the way.

This fixes issue #50


NoahTheDuke (approved):
This looks good to me!

PR #360, opened 2019-05-31
Be defensive against decks with no mwl set (casual play).

### LINKED ISSUE #348: Error 500 when downloading deck as a "Text file" for certain decks
I get 500 when i try do download the following deck - https://netrunnerdb.com/en/decklist/54886/stabby-foxx-1st-at-bristol-sc as a "Text file". 

I DO NOT get the error when doing the same for https://netrunnerdb.com/en/decklist/55245/stabby-maxx-4th-gent-gnk

plural:
A NULL mwl_id in the deck or dceklist table was previously leading
to a crash like:
  Call to a member function getCards() on null

I need to follow this up with some refactoring to share this code between the two controllers.  The way ti is set up and duplicated now is silly.

Fixes #348

### DIFF
--- a/src/AppBundle/Controller/BuilderController.php
+++ b/src/AppBundle/Controller/BuilderController.php
@@ -438,7 +438,10 @@ class BuilderController extends Controller
             );
         }
 
-        $mwl = $deck->getMWL()->getCards();
+        $mwl = null;
+	if ($deck->getMWL()) {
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Dang, what a weird design. Thanks for catching this!

NoahTheDuke (approved):

x3N1GM4x:
Deployed to Production.

PR #361, opened 2019-06-03
Retrieve deck size properly in the decklist text export code path.

### LINKED ISSUE #348: Error 500 when downloading deck as a "Text file" for certain decks

plural:
decklist objects don't have a deckSize() method, like a deck does.

Should really fix #348.

### DIFF
--- a/src/AppBundle/Controller/SocialController.php
+++ b/src/AppBundle/Controller/SocialController.php
@@ -739,7 +739,7 @@ class SocialController extends Controller
         );
 
         if ($decklist->getSide()->getCode() == "corp") {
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Should we just add a `getDeckSize()` function to the model?

plural:
we could do that, but i sort of want to wait until I have a chance to take a closer look at what deck and decklist should have for common interfaces.  I think there is a way to simplify a bunch of this, but i need to read a bit more.

Issue #280, opened 2018-12-23
Commenting on Card Reviews


ghost:
When commenting a card review it would be nice when the # whatever thing would work to supporting linking of cards. Like it does when you comment a deck description.
It surely is an enhancement and has no priority.

plural:
@5N00P1 this should be fixed now.  Please re-open if you have any trouble with it.

PR #385, opened 2019-12-28
Add markdown support for comments, including a preview box.

plural:
This addresses enhancement request #280 

Published version:
Screenshot 2019-12-28 at 3 20 40 PM

Preview while editing.
Screenshot 2019-12-28 at 3 21 13 PM

Reviews have both rawtext and text fields, but I didn't do that here.  I certainly can if we think it is worth it.  This leaves existing comments alone and only new comments will have this functionality.


NoahTheDuke (approved):

x3N1GM4x:
Deployed to production.

Issue #180, opened 2017-12-20
Writing Deck reviews


ghost:
When you write a deck review only cards from the same side are supported automatically with #Cardname so as a corp this only works for corp card and you can't link that easy to runner cards e.g. like DNA Tracker as the runner.
This works well in card reviews and I assume is not that hard to implement.

x3N1GM4x:
I'm not sure what you mean by a "deck review", but it's certainly possible to tag cards from both sides in a comment on a published deck using `#` if that's what you mean?
![image](https://user-images.githubusercontent.com/10621139/49655488-b2ceb480-fa32-11e8-9468-8a5a32f94880.png)
Or do you mean in the description of a deck when you publish it? I think all you can do there currently is insert links with markdown, the `#` prefix doesn't seem to do anything there right now.

ghost:
It works in the comments, but when you write it in Notes, it don't.
![nrdb_notes](https://user-images.githubusercontent.com/5659287/49684299-89fdfc00-fad2-11e8-8d9e-746664e5e369.png)

Create what ever deck go to Notes (which is used when publishing) and try to reference a corp card in a runner deck.

locks:
<img width="1102" height="413" alt="Image" src="https://github.com/user-attachments/assets/270f981a-dfb2-47eb-9e54-9a2c45f952c4" />

Issue is now resolved!

PR #386, opened 2019-12-29
Allow auto-completion of all cards in deck notes/descriptions

plural:
This addresses issue #180 .

The cards you can interact with on the deckbuilder are still locked to the side for the ID, but the notes box now lets you complete all cards.

Notes editing with cards from both sides


NoahTheDuke (approved):

x3N1GM4x:
Deployed to production.

Issue #179, opened 2017-12-20
Updating a deck description


ghost:
When you want to update your deck description a small box on the screan appears which is totally not sufficient to update the deck description.
* It's to small
* It doesn't support the card reference with #Cardname so you need to include it manually

x3N1GM4x:
With regards to the size of that text box, you can drag it to be larger with the control in the bottom right corner:
![image](https://user-images.githubusercontent.com/10621139/49655793-89625880-fa33-11e8-9ed0-05f8914da1be.png)
As for adding `#card-name` support to that text box, is that different you what you've logged under #180?

ghost:
Yes it is, as it's not working at all!
You can type "#" and whatever, but nothing is happening here. See Screenshot

![nrdb_edit](https://user-images.githubusercontent.com/5659287/49684279-31c6fa00-fad2-11e8-855b-97ff9d05b9a4.png)

x3N1GM4x:
Ok, I'm with you, so it's just the #card-name functionality that remains to be resolved on this, you're happy that the field on the form can be resized?

ghost:
Totally!

plural:
This is fixed now with the new publish deck flows.

PR #387, opened 2019-12-29
Add a preview to the publish deck flows

plural:
This addresses issue #179 

Publishing a decklist now has the same preview functionality as other parts of the site including symbol replacement with $ and card replacement with #.  This should make it easier to see exactly what you are going to publish.

Before:
publish decklist pop-up

After:
Screenshot 2019-12-28 at 7 20 44 PM
Screenshot 2019-12-28 at 7 26 26 PM

### DIFF
--- a/app/Resources/views/Builder/decks.html.twig
+++ b/app/Resources/views/Builder/decks.html.twig
@@ -312,11 +312,10 @@ var Identity = null,
 								<label for="publish-deck-description">Description</label>
 								<textarea class="form-control" name="description" id="publish-deck-description" rows="5" placeholder="Enter a brief explanation of the deck strategy and your significant choices"></textarea>
 							</div>
-<div>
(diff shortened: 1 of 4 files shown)


app/Resources/views/Builder/decks.html.twig
```diff
@@ -1,3 +1,4 @@
+
```
NoahTheDuke, line 1: 👀

web/js/deckview.js
```diff
@@ -1,4 +1,5 @@
 
+
```
NoahTheDuke, line 2: 👀

NoahTheDuke (approved):

x3N1GM4x:
Deployed to production. We can fix any formatting niggles separately. :)

app/Resources/views/Builder/decks.html.twig
plural: oops!

PR #388, opened 2019-12-29
Fix published decklist comparison functionality from the decklist page.

### LINKED ISSUE #279: Compare with another Decklist not working any more
I think this issue exists since a longer time, but I never found the time to report it.
When you go to a published deck there is a function to compare it to another one.
Actions -> Compare (2nd entry)
When you select it, nothing happens.
nrdb_compare

plural:
The menu item just wasn't wired up to the javascript to bring up the UI for the comparison.

This fixes #279 

Dialog after clicking "Compare with another decklist"
decklist-comparison-dialog

Decklist Comparison View
decklist-comparison-diff


NoahTheDuke (approved):
Wow, this would be funny if it weren't so annoying. Thanks for figuring it out!

x3N1GM4x:
Deployed to production.

PR #391, opened 2019-12-30
Card completion everywhere!!!!

plural:
This has a bunch of changes.  Sadly a bunch of duplication, but i'll tidy that up in a follow-up.

The top nav now has auto-completion, but it is in a drop down.  There is the same number of clicks right now, but i'm definitely looking for feedback there.

The card page needs to put the code for the card into the search box in order to get to the card you actually selected.  Otherwise cards whose names are substring matches for other cards will show the search results page, not the card you selected in the dropdown.

# Top Nav
top-nav-search-1
top-nav-search-2
top-nav-search-3
top-nav-search-4

# Card Page
card-page-autocomplete-1
card-page-autocomplete-2
card-page-autocomplete-3

# Advanced Search Page
advanced-search-autocomplete-2
advanced-search-autocomplete-3

### DIFF
--- a/app/Resources/views/Search/advanced-search.html.twig
+++ b/app/Resources/views/Search/advanced-search.html.twig
@@ -6,6 +6,45 @@
 <div class="container">
 
     <script type="text/javascript">
+        Promise.all([NRDB.data.promise, NRDB.ui.promise]).then(function() {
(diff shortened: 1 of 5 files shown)


NoahTheDuke (approved):

x3N1GM4x:
Deployed to production.

PR #392, opened 2019-12-30
Rename 'Spoilers' on the card search form to be 'Text only' 

plural:
This is just a tiny fix to make things more logical.

spoilers-1
spoilers-2


NoahTheDuke (approved):

x3N1GM4x:
Deployed to production.

PR #395, opened 2020-01-03
Reduce code duplication for the deck publishing and decklist editing dialogs

plural:
This extracts a common html template and common javascript to reduce duplication across the 3 places this is used.


NoahTheDuke:
Looks good to me!

x3N1GM4x:
Deployed to production.

Issue #353, opened 2019-05-12
Add link to ABR


ghost:
ABR (https://netrunnerdb.com/en/profile/18163/5N00P1) has a nice profile page (e.g. https://alwaysberunning.net/profile/18163) with a field where you can enter your NRDB User name and it will create a link to your NRDB profile page (e.g. https://netrunnerdb.com/en/profile/18163/5N00P1)
It would be nice when the profile page of NRDB would have the same linking back to ABR.
The ideal implementation would be a new field in the account page where you can enter your username and it will appear with link to the profile page of ABR.
The ABR implementation is easy and slim.
It should be optional so the people sign actively in.

NoahTheDuke:
This is a great idea! I’ll look into it.

plural:
There is now a link to the ABR profile if one exists (see #399)

Issue #394, opened 2019-12-31
Field to ABR Username


ghost:
let people link their profile to ABR by simply adding a field and make it possible to find their ABR profile. In ABR it's just a number and it's hard to find users. And this way people can do it, but don't have to. Like it's already in ABR

plural:
does ABR even have a separate username?  it offers login through NRDB and the user id is the same as the NRDB username. 

Adding a link to the profile page for the ABR page would be easy and really nice, but I'm not sure I understand this request as is.

PR #397, opened 2020-01-05
Add a link to a user's Always Be Running Profile in the user dropdown.

plural:
Since ABR does logins via NRDB the URL is knowable trivially.

There may be a better place for this, but this is a good start.

I think this addresses the spirit of #394


plural:
Screenshot of the profile page change: 
Screenshot 2020-01-05 at 12 48 24 AM

This could be soooooooo much better, but I think it is significantly better than nothing.

plural:
Fixes #353

x3N1GM4x:
Deployed to production.

PR #398, opened 2020-01-05
Add a secondary sort by decklist name to the decklist search page to …

### LINKED ISSUE #316: Published Decks by User has inconsistent paging
I have 53 published decks. 
https://netrunnerdb.com/en/decklists/find?author=5N00P1
so more then one page is used for this. But when going to page 2, I might see the same decks as on page 1 and going back to page 1 it might look different then before. So it's hard to find some decks ;-)

I would say, the sorting algorithm is not reliable enough to always produce the same results. Which for me when looking for a deck and recomend it to some one is an issue.
Please see screenshots from page 1 before & after reload.
snap1
snap2

plural:
…make paging deterministic.

This fixes #316

### DIFF
--- a/src/AppBundle/Service/DecklistManager.php
+++ b/src/AppBundle/Service/DecklistManager.php
@@ -824,7 +824,7 @@ class DecklistManager
                 where $where
                 and d.moderation_status in (0,1)
                 $group_by
-                order by $order desc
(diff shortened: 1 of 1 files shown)


x3N1GM4x:
Deployed to production.

PR #403, opened 2020-01-15
Take the user to the card page when choosing an item on the card page search box.

plural:
make this behavior match the top nav search behavior.


x3N1GM4x:
LGTM and tested fine on my local, nice one!

x3N1GM4x:
Deployed to production.

PR #404, opened 2020-01-15
Lighthouse audit fixes

plural:
i ran a report using chrome's lighthouse auditing tool.  It identified some outdated javascript deps with security risks.  There were also some deps that we weren't fetching minimized versions of.
Fetch some of the resources asynchronously instead of blocking.


plural:
this audit also spurred turning on gzip compression on the server and enabling the HTTP2 protocol.

x3N1GM4x:
Deployed to production.

Issue #409, opened 2020-01-20
Spaces missing in drop down menus


Tamijo-Netrunner:
Clicking on various drop down menus throughout the site seems to have missing spaces between words.

Seen here is the drop down for "sort by" in searches, but this also occurs on decklist screens for rotation legality.

![Screenshot_20200117-123426_Chrome](https://user-images.githubusercontent.com/39098450/72741876-71e72000-3ba0-11ea-8674-abf9cb06e38e.jpg)

plural:
should be fixed now.

Issue #406, opened 2020-01-20
Creating new Decks, Faction Symbols are missing


ghost:
When creating a new deck, the faction symbol is not longer visible besides the ID name, like it was in the past.
![image](https://user-images.githubusercontent.com/5659287/72724373-6df8d500-3b83-11ea-8ef6-72133735cef3.png)

plural:
should be fixed now

PR #410, opened 2020-01-22
Update doctrine/orm and synchronize composer.json and lock.

plural:
Updated with composer update.  Manually bumped doctrine/orm to get ride of the UnitOfWork errors.

This also fixes the spacing issues we've been seeing.


plural:
Fixes #409 
Fixes #406

Issue #309, opened 2019-01-27
Add a link in Advanced Search (or similar) to "standard" legal cards


AndersRask:
I have seen quite a few in slack and Dorks asking for a filter for cards legal in current Standard format.
I did the work for them giving them the below link, but if this could be nice with having this handy as a link or filter (checkbox next to search) or adv search
https://netrunnerdb.com/find/?q=c%210%7C1%7C2%7C4%7C13%7C20

NoahTheDuke:
Thanks for the idea! Something like this has been discussed internally but I’ve not had a chance to do it yet. I hope to get to it soon tho! I’ll keep you posted with my progress.

ryowufei:
Hi, I'm very much supporting this.

If I may, I would described my view on that feature a bit more:
- A search result to display rotated cards (so that it's simple to pull them out of our physical card pool without having to cross reference with core set 2019 for instance).
- A search result to display what is in a MWL (be able to see/hide them, sorted by status : restricted/banned)
- A search parameter to exclude anything not legal in a search (rotated, banned) depending on chosen rotation and chosen mwl. Or at least a flag in the result list to easily spot them.
- Last, it would be great to see in the deck builder which cards are rotated rather than having a message stating that at least one is not compliant.

Best,

opaldes:
I think it would be easy to add standard as an query condition, as rotations are already inside the card repository. Ill try setting up a local NRDB instance and test it.

My Idea:
1. Add a method to the Rotations service to get the current packs from rotation as array.
2. Add a new "standard" filter to the CardData service for the current rotation

PR #414, opened 2020-02-04
add a z:current filter option

opaldes:
#309

### DIFF
--- a/src/AppBundle/Service/RotationService.php
+++ b/src/AppBundle/Service/RotationService.php
@@ -22,6 +22,22 @@ class RotationService
         $this->entityManager = $entityManager;
     }
 
+    /**
+     * [returns the first entry of the descending sorted rotations]
+     * @return Rotation current Rotation
+     */
+    public function findCurrentRotation()
+    {
+        $rotations = $this->entityManager->getRepository(Rotation::class)->findBy([], ['dateStart' => 'DESC']);
+
+        //there should always be a rotation available
+        if(count($rotations) <= 0){
+            throw new \Exception("No current rotation found", 1);
+        }
+
+        return $rotations[0];
+    }
+
     public function findCompatibleRotation(Decklist $decklist)
     {
         $rotations = $this->entityManager->getRepository(Rotation::class)->findBy([], ['dateStart' => 'DESC']);
--- a/src/AppBundle/Service/CardsData.php
+++ b/src/AppBundle/Service/CardsData.php
@@ -8,6 +8,8 @@ use AppBundle\Entity\Mwl;
 use AppBundle\Entity\Pack;
 use AppBundle\Entity\Review;
 use AppBundle\Entity\Ruling;
+use AppBundle\Entity\Rotation;
+use AppBundle\Service\RotationService;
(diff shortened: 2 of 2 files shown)


NoahTheDuke:
Thanks for this! I'll test it out and see!

src/AppBundle/Service/RotationService.php
```diff
@@ -22,6 +22,22 @@ public function __construct(EntityManagerInterface $entityManager)
         $this->entityManager = $entityManager;
     }
 
+    /**
+     * [returns the first entry of the descending sorted rotations]
+     * @return Rotation current Rotation
+     */
+    public function findCurrentRotation()
+    {
+        $rotations = $this->entityManager->getRepository(Rotation::class)->findBy([], ['dateStart' => 'DESC']);
```
plural, line 31: you could also consider

$rotation = $this->entityManager->getRepository(Rotation::class)->findOneBy([], ['dateStart' => 'DESC']);
if (!$rotation) {
  throw ...
}
return $rotation;

plural (changes requested):
This is excellent!  I patched it in and it works well.

Please also update the syntax page at app/Resources/views/Default/syntax.html.twig and the tooltip for syntax at app/Resources/views/Search/searchtooltip.html.twig

(i should pull those together so we don't have them duplicated...)

Another option we could consider here is expanding this just a little bit:

z:rotation#|current|latest

the rotations have names and codes in the DB right now, but AFAIK, they aren't shown anywhere.

mysql> SELECT * FROM rotation;
+----+---------------+-----------------+---------------------+---------------------+------------+
| id | code          | name            | date_creation       | date_update         | date_start |
+----+---------------+-----------------+---------------------+---------------------+------------+
|  1 | rotation-2017 | First Rotation  | 2019-12-26 22:29:02 | 2019-12-26 22:29:02 | 2017-10-01 |
|  2 | rotation-2018 | Second Rotation | 2019-12-26 22:29:02 | 2019-12-26 22:29:02 | 2018-12-21 |
|  3 | rotation-2019 | Third Rotation  | 2019-12-26 22:29:02 | 2019-12-26 22:29:02 | 2019-12-27 |
+----+---------------+-----------------+---------------------+---------------------+------------+

src/AppBundle/Service/RotationService.php
opaldes: will do seems better because we expect only 1 to return

jeremy-seipelt: @plural do we want z:"First Rotation" or z:rotation-2017 ? I would change my function appropriately

plural: no quoting is better, i think, so i would go with rotation-2017.  We can always publish the names of the valid rotation strings on the syntax page to help people out, too.

plural:
hey @opaldes, do you want to push ahead with this or have one of us run with it?  Thanks for kicking this off.

plural:
i ran with this in #427 and merged it in.  Thanks again to @opaldes and @jeremy-seipelt for kicking this off!

Issue #413, opened 2020-02-02
Add a Banned icon for cards banned in various formats


plural:
This should probably be present on the deck view and deck list pages as well as in the card search.

It is probably safe to default to the current legality when looking at card search or set views.

Deck building should probably share more info about what cards are banned or restricted in which formats to make the "issues" super clear to anyone reading the decks.

PR #415, opened 2020-02-04
add a prohibited emoji to the unicorn function

opaldes:
#413

### DIFF
--- a/web/js/nrdb.js
+++ b/web/js/nrdb.js
@@ -257,10 +257,24 @@ function find_identity() {
     Identity = NRDB.data.cards.find({ indeck: { '$gt': 0 }, type_code: 'identity' }).pop();
 }
 
+/**
+ *  takes a card and adds an unicorn based on the current card
+ * @param  card
+ * @return string
+ */
 function unicorn(card) {
+    var emoji = ""
     var mwlCard = get_mwl_modified_card(card);
     var unicorn_emoji = ' <span title="Restricted card" style="display:inline-block;width:1.5em;">🦄</span> ';
-    return mwlCard.is_restricted ? unicorn_emoji : '';
+    var prohibited_emoji = ' <span title="Restricted card" style="display:inline-block;width:1.5em;">🚫</span> ';
+
+    if (mwlCard.is_restricted) {
+        emoji = unicorn_emoji;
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Thanks for this! I'll test it out and hopefully we can get it live soon.

web/js/nrdb.js
```diff
@@ -257,10 +257,24 @@ function find_identity() {
     Identity = NRDB.data.cards.find({ indeck: { '$gt': 0 }, type_code: 'identity' }).pop();
 }
 
+/**
+ *  takes a card and adds an unicorn based on the current card
```
plural, line 261: how about "Returns a banned or restricted icon for the supplied card and selected MWL." ?

web/js/nrdb.js
```diff
@@ -257,10 +257,24 @@ function find_identity() {
...
+ * @param  card
+ * @return string
+ */
 function unicorn(card) {
+    var emoji = ""
     var mwlCard = get_mwl_modified_card(card);
     var unicorn_emoji = ' <span title="Restricted card" style="display:inline-block;width:1.5em;">🦄</span> ';
-    return mwlCard.is_restricted ? unicorn_emoji : '';
+    var prohibited_emoji = ' <span title="Restricted card" style="display:inline-block;width:1.5em;">🚫</span> ';
```
plural, line 269: not strictly a necessary change, but since you only use each of these once, you could just move them into the conditional block and remove the other variables.

if (mwlCard.is_restricted) {
  return "blah blah"
} else if (mwlCard.deck_limit == 0) {
  return "blah blah"
}
return "";

plural (changes requested):
nice, thanks!  This looks great.  I don't know of any other places that need this off the top of my head, but we'll find them when we release it if we missed any.  :D

web/js/nrdb.js
opaldes: Seems reasonable

plural:
hey @opaldes this is great.  Did you want to pick this back up or have one of us run with it?  This will be a really nice improvement for the UI!

plural:
big thanks to @opaldes for getting this kicked off.  I ran with this in #426 because we didn't hear back for a bit.

PR #418, opened 2020-02-15
Properly update the collection filters when using the collection shortcuts

### LINKED ISSUE #417: Collection pack select shortcuts not updating card lists
It seems using the shortcuts (Current Cardpool | All | None) don't update the card lists. After selecting one of them (such as Current Cardpool), selecting and deselecting a single pack updates the lists for builder.

plural:
Sorry these didn't work well originally!

This fixes #417 

Screenshot 2020-02-15 at 12 16 20 PM
Screenshot 2020-02-15 at 12 16 35 PM
Screenshot 2020-02-15 at 12 16 48 PM
Screenshot 2020-02-15 at 12 17 02 PM
Screenshot 2020-02-15 at 12 17 11 PM

There is more that could be done to tidy this up.  I think there are a couple of other deckbuilder bugs that probably share a similar root cause, but i'll look at those separately.

### DIFF
--- a/web/js/deck.js
+++ b/web/js/deck.js
@@ -189,18 +189,21 @@ function create_collection_tab(initialPackSelection) {
 		$('#pack_code').find(':checkbox').each(function() {
 			$(this).prop('checked', !(rotated_cycles[$(this).prop('name')] || rotated_packs[$(this).prop('name')]));
 		});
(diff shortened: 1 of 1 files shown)


NoahTheDuke (approved):
Kind of awkward, but the best we currently have.

PR #424, opened 2020-03-29
Removed redundant pagination controls

TylerMorley:
Fixed bug introduced in Alsciende#419 where cycles would have two sets of pagination buttons.


plural:
hey there! Thanks for sending this in!

I patched it in and it looks great. This fixes things for cycles.  Here are some before/after shots:

Checklist Before:
cycle-checklist-nav-before

Checklist After:
cycle-checklist-nav-after


Full Cards Before:
cycle-nav-full-cards-before

Full Cards After:
cycle-nav-full-cards-after

PR #426, opened 2020-04-07
add banned icon to the unicorn function.

plural:
Copy over https://github.com/Alsciende/netrunnerdb/pull/415 and incorporate my code review comments since we didn't hear back on the review.  Big thanks to @opaldes for the initial pull request.

new version of #415


NoahTheDuke (approved):
Looks good to me!

plural:
Note that this doesn't yet show up on the deck page for deck ids for #reasons.  Basically, this is all JS side stuff and that stuff is all server side .... mostly.  I'll file a different bug and fix it later.

PR #427, opened 2020-04-07
Add search operator to specify rotations.

plural:
This is based off of https://github.com/Alsciende/netrunnerdb/pull/414 by @opaldes.

I incorporated my feedback on the original PR and updated the syntax pages as well.

I picked this up because we haven't heard back for a while.

Partially addresses #309 

(note that we already de-dupe to the latest card when multiples are present, so yo don't see some of the cards in the original core set)
z-op-current-data

# Runner cards, 2017 rotation
z-op-runner-2017

# Runner cards, 2018 rotation
z-op-runner-2018

# Runner cards, 2019 rotation
z-op-runner-2019

### DIFF
--- a/src/AppBundle/Service/CardsData.php
+++ b/src/AppBundle/Service/CardsData.php
@@ -8,6 +8,8 @@ use AppBundle\Entity\Mwl;
 use AppBundle\Entity\Pack;
 use AppBundle\Entity\Review;
 use AppBundle\Entity\Ruling;
+use AppBundle\Entity\Rotation;
+use AppBundle\Service\RotationService;
 use AppBundle\Repository\PackRepository;
 use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Component\Asset\Packages;
@@ -166,7 +168,6 @@ class CardsData
 
         $qb2 = null;
         $qb3 = null;
-
         $clauses = [];
         $parameters = [];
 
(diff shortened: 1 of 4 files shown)


src/AppBundle/Service/CardsData.php
```diff
@@ -580,6 +581,22 @@ public function get_search_rows(array $conditions, string $sortorder, string $lo
                     }
                     $i++;
                     break;
+                case 'z': // rotation
+                    // Instantiate the service only when its needed.
+                    $rotationservice = new RotationService($this->entityManager);
+				    $rotation = null;
```
NoahTheDuke: 👀

src/AppBundle/Service/CardsData.php
```diff
@@ -580,6 +581,22 @@ public function get_search_rows(array $conditions, string $sortorder, string $lo
...
+                    if ($condition[0] == "current" || $condition[0] == "latest") {
+                        $rotation = $rotationservice->findCurrentRotation();
+                    } else {
+						$rotation = $rotationservice->findRotationByCode($condition[0]);
```
NoahTheDuke: looks like these are tabs and not spaces? You'll wanna normalize to spaces.

src/AppBundle/Service/CardsData.php
```diff
@@ -580,6 +581,22 @@ public function get_search_rows(array $conditions, string $sortorder, string $lo
...
+                    }
+                    if ($rotation) {
+                        // Add the valid cycles for the requested rotation and add them to the WHERE clause for the query.
+                        $cycles = "'" . implode("','", $rotation->normalize()["cycles"]) ."'";
+                        $clauses[] = "(y.code in ($cycles))";
```
NoahTheDuke: I think this should be `(y.code in (?$i))` and then `$parameters[i++] = $cycles;`, right?

src/AppBundle/Service/CardsData.php
plural: oops.  Fixed.

src/AppBundle/Service/CardsData.php
plural: yep, sorry.  Fixed.

src/AppBundle/Service/CardsData.php
plural: i took a slightly different path but also ended up here.  Thanks for pointing out the parameter stuff.

plural:
cool, thanks @NoahTheDuke !  This is in better shape now.

PR #429, opened 2020-04-11
Lazily load (most) images

plural:
This is another improvement suggested by the lighthouse audit.  This should help out page responsiveness.


NoahTheDuke:
This is very cool. Won’t have much of an effect until the browsers support it (only Firefox at the moment), but will be great after that happens.

plural:
hmm, i'm not seeing what isn't supported by firefox here, but i'll check on firefox directly in a bit.

NoahTheDuke:
My apologies, my understanding of the support was wrong (and I phrased it backwards lol). Looks like the modern browsers except Safari support it: [here](https://www.caniuse.com/#search=loading)

plural:
gotcha.  I'll double-check on a safari browser in a bit, as well.  I think
this lib is just using JS for it and not using that attribute, so we should
be good, but i will verify.

-jason

app/Resources/views/Decklist/decklist-row-always-xs.html.twig.orig
```diff
@@ -0,0 +1,57 @@
+<<<<<<< HEAD
```
NoahTheDuke: uh oh, looks like a merge conflict snuck in

plural: oops!  Looks like i let 2 files accidentally get checked in from a merge conflict.

plural:
thanks for taking another look, Noah.

app/Resources/views/Decklist/decklist-row-always-xs.html.twig
```diff
@@ -1,6 +1,6 @@
 <tr>
   <td style="width:67px">
-    <img src="{{ decklist.identity_url ? decklist.identity_url : asset('/card_image/'~decklist.code~'.png') }}" alt="{{ decklist.identity }}" style="width:51px;height:71px">
+    <img data-src="{{ decklist.identity_url ? decklist.identity_url : asset('/card_image/'~decklist.code~'.png') }}" alt="{{ decklist.identity }}" style="width:51px;height:71px" class="lazyload">
```
NoahTheDuke, line 3: `src` is a [required attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img) of `img` tags. Did you mean to add an additional `data-src` tag?

plural: no, the way this works is by taking advantage of the fact that a missing src is basically a no-op.  With a src, the browser makes the request immediately.  With data-src, the js will do it when appropriate.

PR #430, opened 2020-04-11
Set all files to have unix newlines.

plural:
no other changes.  I just opened these up in vim, used :set fileformat=unix and then saved.


plural:
actually, you don't need to review this, Noah.  I'll just merge this in.

NoahTheDuke:
You can set this in the git attributes file as well! Will force it for all files

plural:
yeah, gonna figure out what we need to do to keep this good going forward.
Expect an equally annoying CL for tabs in a bit here, too.  :)

-jason

PR #432, opened 2020-05-04
Use new /card_image/{size}/code.jpg url patterns for card images.

plural:
This will send an appropriately sized jpegs out with page requests
instead of the full size png files every time.

This shaves a ton of data off each request.

we won't be able to push this live until i wrap up the scripts to make the images look good at all these sizes and hopefully have nice rounded corners as well.


app/Resources/views/Search/display-images.html.twig
```diff
@@ -9,7 +9,7 @@
 {% if index in cards|keys %}
 {% set card = cards[index] %}
 <div class="col-sm-3" style="margin-bottom:30px">
-  <a href="{{ card.url }}" class="no-popup"><img src="{{ card.imageUrl }}" alt="{{ card.title }}" style="width:100%" class="img-responsive"></a>
+  <a href="{{ card.url }}" class="no-popup"><img src="{{ asset('/card_image/large/'~card.code~'.jpg') }}" alt="{{ card.title }}" style="width:100%" class="img-responsive"></a>
```
NoahTheDuke: Should this be medium sized?

app/Resources/views/Faction/faction.html.twig
```diff
@@ -23,7 +23,7 @@
 
 <div class="row">
 <div class="col-md-3">
-<a href="{{ path('cards_zoom',{_locale:app.request.locale,card_code:item['identity'].code}) }}" class="no-popup"><img src="{{ item['identity'].imageUrl ? item['identity'].imageUrl : asset('/card_image/'~item['identity'].code~'.png') }}" class="img-responsive"></a>
+<a href="{{ path('cards_zoom',{_locale:app.request.locale,card_code:item['identity'].code}) }}" class="no-popup"><img src="{{ asset('/card_image/large/'~item['identity'].code~'.jpg') }}" class="img-responsive"></a>
```
NoahTheDuke: Should this be medium?

app/Resources/views/Search/display-images.html.twig
plural: dimensions i pulled for the sizes based on how things are used / displayed right now are:

tiny: 51x71
small: 116x162
medium: 165x230
large: 300x419

these images are 263x363 as displayed on the page right now.

app/Resources/views/Faction/faction.html.twig
plural: these are currently displayed as 263x366, so large (see sizing below) for the best fit without upscaling.

NoahTheDuke:
This looks great. :shipit:

plural:
i made one pass through to add a helper method.  This leaves off JS accesses and images that come from SQL queries for things like deck lists.  This is better, though.  I would prefer to make a single pull request that makes the JS image access nice and clean.

PR #452, opened 2020-10-29
Fix banlist for decklist search.

plural:
The ban list drop down has the mwl code (string form like standard-ban-list-20-06) and the query had the right SQL, but the type for the parameter was set to INT, causing the param to be dropped.  I guess it's nice that this just silently failed instead of crashing, but it definitely wasn't working.  :)


NoahTheDuke (approved):
Oh dang, nicely caught!

PR #453, opened 2020-11-04
Update NRDB to use Nisei logo

plural:
It's time for a small update for NetrunnerDB.  It should also have the NISEI logo!

Screenshot 2020-11-03 at 9 44 38 PM


plural:
we discussed this on the web dev slack and are going to do a bit of a refresh, but not like this.   Closing this out.

PR #457, opened 2020-12-15
Add a button in the bottom section of the site for our patreon.

plural:
this could be prettier, but it's fine for now.


NoahTheDuke (approved):
Much better than on the header.

PR #460, opened 2020-12-23
Finally check in a tiny image loading fix i had deployed already.

plural:
I need to do a bit more work on optimizing site performance, but this
one was not good and caused too much flashing.

I had done this manually a while back and forgot to check it in.


NoahTheDuke (approved):
Idk anything about this so I trust you lol

PR #474, opened 2021-03-02
Add the script used to summarize the releases.

plural:
This requires the command line github utilities and for the user to log in with their credentials in order to publish a draft release.

I'm pretty sure I got this from Noah's recommendation at some point.


NoahTheDuke (approved):
Looks good to me!

PR #475, opened 2021-03-03
Update Card page to look better

NoahTheDuke:
image
image


plural (approved):
I'd be interested in seeing this with the card image centered, but THIS IS SO LOVELY.  :)

I'll leave it to you to submit if you care to change it to do that or not.

PR #477, opened 2021-03-07
Sort the sets page from newest to oldest.

plural:
Let's keep the hype going!  New cards are awesome, old cards are necessary!

I am a little surprised but this function is only used in one place, so this appears to be all we need.

Screenshot 2021-03-07 at 10 02 43 AM


NoahTheDuke (approved):
perfection

PR #481, opened 2021-03-13
Add strippedTitle

NoahTheDuke:
Follow-up to #469. Paired with [netrunner-cards-json#557](https://github.com/NetrunnerDB/netrunner-cards-json/pull/557).

### DIFF
--- a/src/AppBundle/Command/ImportStdCommand.php
+++ b/src/AppBundle/Command/ImportStdCommand.php
@@ -33,39 +33,6 @@ class ImportStdCommand extends ContainerAwareCommand
 
     private $collections = [];
 
-    private $xmlToStrip = [
-        "/<strong>/",
-        "/<\/strong>/",
-        "/<trace>/",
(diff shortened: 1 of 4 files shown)


src/AppBundle/Command/ImportStdCommand.php
```diff
@@ -491,12 +459,6 @@ protected function copyFieldValueToEntity($entity, string $entityName, string $f
                         }
                 }
             }
-        } elseif ($fieldName == "strippedText") {
```
plural, line 494: are you ready to remove this in this pull request?  You aren't pulling stripped_text in from the json yet.

NoahTheDuke: Line 332 notes it as an optional field.

plural: ahh, gotcha.  it was hidden in the review by default.

plural (approved):

PR #485, opened 2021-03-14
Move (show history) onClick to js file

NoahTheDuke:


web/js/zoom.js
```diff
@@ -349,3 +350,13 @@ function write_review_open(event) {
     }
 
 }
+
+function show_mwl_history(event) {
+    event.preventDefault();
+
+    let entries = $.find('tr.card-mwl-inactive');
+    entries.forEach(x => x.style.display = (x.style.display == 'table-row' ? 'none' : 'table-row'));
+    let l = $('#mwl-history');
```
plural: i think you can do event.target and skip the extra variable.  Up to you.

plural (approved):
one optional thing.

PR #486, opened 2021-03-14
Basic startup support

plural:
Sort the packs in the deckbuilder newest -> oldest.
Change pack selection to Startup | Standard | All | None
Screenshot 2021-03-13 at 7 00 04 PM

 Add a simple Startup Card Pool indicator to the card page.
Screenshot 2021-03-13 at 6 21 45 PM
Screenshot 2021-03-13 at 6 20 42 PM

### DIFF
--- a/src/AppBundle/Controller/SearchController.php
+++ b/src/AppBundle/Controller/SearchController.php
@@ -469,7 +469,6 @@ class SearchController extends Controller
                 $cardinfo['available'] = $availability[$pack->getCode()];
 
                 $currentRotationCycles = [];
-
                 if ($view == "zoom") {
                     $cardVersions = $versions[$card->getTitle()];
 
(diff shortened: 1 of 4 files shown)


src/AppBundle/Controller/SearchController.php
```diff
@@ -480,6 +480,11 @@ public function displayAction(
                     $cardinfo['versions'] = [];
                     $standard_legal = true;
                     $all_versions_rotated = true;
+
+                    // Startup legality is currently hard-coded since the DB doesn't know anything about it.
+                    $startupCycles = ['ashes' => true, 'system-gateway' => true, 'system-update-2021' => true];
```
NoahTheDuke, line 485: This can be a normal array, and then instead of `array_key_exists`, you use `in_array`

NoahTheDuke (approved):
looks good to me

PR #490, opened 2021-03-22
Tidy up advancement/points indicators for agendas.

plural:
This makes things consistent with the card hover view and common parlance of $advancementCost/$points

Screenshot 2021-03-21 8 16 27 PM
Screenshot 2021-03-21 8 14 46 PM


app/Resources/views/Search/display-zoom.html.twig
```diff
@@ -29,8 +29,7 @@
                     </span>
                     <span class="card-cost">
                         {% if card.type_code == "agenda" %}
-                            {{ card.advancementcost }} <span class="icon icon-rez"></span>
-                            {{ card.agendapoints }} <span class="icon icon-agenda"></span>
+                            {{ card.advancementcost }}/{{ card.agendapoints }}
```
NoahTheDuke, line 32: We use icons for the other numbers here (cost mainly). Any reason to use the slash instead of fixing icon display?

plural: we don't have icons for these in the font at all.  We'd need to get those in as well.  i did go look at the svg to see if we already had them and just hadn't put them in the css yet.

is there a general icon for advancement?  The agenda symbol for points could be good.

NoahTheDuke: We don't? That's wild, I was sure we did. I think this change is good, but I'd like to include the icons in the long run.

plural: i got an icon for agenda points from Krembler and will get that added here.  We don't have one for advancement yet.

plural: i'm going to do the icon stuff as a separate PR.

PR #492, opened 2021-03-24
Update php session settings for longer logins

balkamm:
Here we set gc_lifetime and cookie_lifetime to 1 year so that User Logins don't expire too quickly.  The defaults are "When you close your browser window" and 24 minutes (for the server session), which is too short for most users.


plural:
hahahaha

our version of symfony is too old for those options.  I patched this in and the app crashed.  That's our fault, not yours.  :D 

I got this to work locally:
+        # One month = 30 * 24 * 60 * 60
+        gc_maxlifetime: 2592000
+        cookie_lifetime: 2592000

and shortened things to a month for idle time.

balkamm:
lol.  Done. Switched it to 1 month.

plural (approved):

PR #493, opened 2021-03-24
Sort packs newest to oldest.

plural:
oops.  We missed that packs were in oldest to newest order, even after we changed sets to be newest to oldest.  :D


NoahTheDuke (approved):

francescopellegrini:
I'm still seeing packs in oldest-to-newest order in the decklist tab..

image

PR #498, opened 2021-03-27
Don't allow selecting 0 copies of your ID.

### LINKED ISSUE #455: Setting current identity to 0 copies causes card browser to break
Setting a new identity by increasing the count from 0 to 1 of some identity not already selected works fine. However, a new user such as myself may be very likely to try to set the current identity count to 0 first, either via the modal that pops up when clicking the identity at the top of the decklist (despite the cursor changing to a blocked symbol, the 0 button in that button group is still clickable), or through setting the card count of the current identity to 0 in the card browser on the right. When this happens, the website breaks. If cards are showing on the card browser pane at this point, any number of then can be "added" without affecting anything.

Screen Shot 2020-11-24 at 11 08 19

Doing anything else to the state of the card browser causes it to show 0 results, and searching for cards via it no longer causes them to pop up in a modal. 

The bad state can be fixed by saving and reopening the deck, but this issue initially caused me to think it wasn't possible to change the identity of a deck that had already been created.

plural:
Fixes #455 

Don't do anything if you select 0 copies of your ID.  That's silly.

### DIFF
--- a/web/js/deck.js
+++ b/web/js/deck.js
@@ -370,6 +370,23 @@ $(function() {
 
   $('#collection').on({
     change : function(event) {
+      var el = event.target;
+      // don't allow setting your current ID to 0 copies.  That's silly.
+      var index = $(el).closest('.card-container').data('index') || $(el).closest('div.modal').data('index');
+      var card = NRDB.data.cards.findById(index);
+      if (card.type_code == 'identity') {
+        var in_collection = $(el).closest('#collection').length;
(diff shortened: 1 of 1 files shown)


NoahTheDuke (approved):
I trust that this works but wtf deckbuilder.

plural:
hahaha, for sure.

It took a while to figure out where this needed to be to prevent all the stuff from taking effect.

PR #506, opened 2021-03-31
Add an MwlCard entity and update importer for it.

plural:
This makes a new entity with a field for each mwl/banlist value for a card (printing).

This is step 1 of fixing #452

### DIFF
--- a/src/AppBundle/Entity/MwlCard.php
+++ b/src/AppBundle/Entity/MwlCard.php
@@ -1,0 +1,210 @@
+<?php
+
+namespace AppBundle\Entity;
+
+use AppBundle\Behavior\Entity\NormalizableInterface;
+
+/**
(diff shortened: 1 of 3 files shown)


src/AppBundle/Entity/MwlCard.php
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * @var integer
+     */
+    private $mwl_id;
+
+    /**
+     * @var integer
+     */
+    private $card_id;
```
NoahTheDuke: Does this need to be a foreign key?

plural: yeah.  these are unnecessary.  The FK fields were already there and i was already using them in the importer.

NoahTheDuke (approved):

PR #511, opened 2021-04-02
Fork and fix the typeahead code to fix the tab searchbox behavior.

plural:
the typeahead library hasn't been updated in years, but stackoverflow pointed me to the _onTabKeyed function.  This patch has the source (mit license) and a minified version of it.

The behavior here now is that tab is the same as down arrow.


NoahTheDuke (approved):
Any reason to not continue linking to a cdn for the library?

plural:
The CDN version doesn't have the patch to change the tab key behavior that actually deals with the issue.  There are other, more invasive, changes like switching away from the typeahead library, which we can do separately (and probably should consider).

web/js/typeahead.jquery.js
```diff
@@ -1127,8 +1127,11 @@
             },
 			// This fixes the tab behavior for the card search box.
             _onTabKeyed: function onTabKeyed(type, $e) {
-                $e.preventDefault();
                 this.moveCursor(+1);
+                var $selectable;
+                if ($selectable = this.menu.getActiveSelectable()) {
```
NoahTheDuke: Can I ask that we don't assign in if statements?

plural: sure.  tidied up.

PR #512, opened 2021-04-03
Invert rotation logic

gereons:
Revival of https://github.com/NetrunnerDB/netrunnerdb/pull/209

Implements the data change from https://github.com/NetrunnerDB/netrunner-cards-json/pull/575

When this gets merged, the steps to deploy this are as follows:
* update the database schema using `php bin/console doctrine:schema:update --force`
* manually delete the contents of the `rotation_cycle` table
* import the new rotations.json with the usual import command

At least in my test db, this did not create new `rotation_id`s (so that all previous decks still point to the correct rotation), but in any case, do make sure to backup the database before the migration… :)

If this does not work somehow, you’d either have to run the `app:legality:decklists-rotation` command to update existing decklists, or manually fix the `decklist.rotation_id` column to update rotations, the mapping from old to new rotations should be straightforward 1:1.

### DIFF
--- a/src/AppBundle/Command/LegalityDecklistsRotationCommand.php
+++ b/src/AppBundle/Command/LegalityDecklistsRotationCommand.php
@@ -74,7 +74,6 @@ class LegalityDecklistsRotationCommand extends ContainerAwareCommand
                 if ($confirm && $oldId !== $rotation->getId()) {
                     $output->writeln("  updating decklist " . $decklist->getId());
                     $decklist->setRotation($rotation);
-                    $this->entityManager->persist($decklist);
                 }
             }
         }
--- a/src/AppBundle/Command/LegalityDecklistsRotationCommand.php
+++ b/src/AppBundle/Command/LegalityDecklistsRotationCommand.php
@@ -6,6 +6,7 @@ use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use AppBundle\Service\RotationService;
 
 /**
  *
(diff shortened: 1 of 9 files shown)


src/AppBundle/Command/LegalityDecklistsRotationCommand.php
```diff
@@ -51,6 +56,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 AND ny.rotated = 0);";
 
         $this->entityManager->getConnection()->executeQuery($sql);
+
```
plural, line 59: this section is something we should probably revisit.

I want to break away from some of the one-and-only assumptions the app has right now.

If we had a decklist/rotation table we could have all the rotation validity info and not have to worry about guessing as to what rotation_id means and treat this more like banlist and (increasingly) cardpool and the developing "format" info.

plural:
Ooh, i like this!

One biggish thing for decklists requested, so let me know what you think.

we definitely need to test this rollout procedure out well.  We haven't been doing many schema updates, so i expect mistakes.  :D

src/AppBundle/Command/LegalityDecklistsRotationCommand.php
gereons: Oh, absolutely, that's just something that goes way over my head as a total PHP/Symphony/Doctrine noob, so I didn't want to touch this part :)

gereons:
Re: the rollout part: I've never had problems with the schema update way back when alsciende made to on a regular basis, and I've run this particular migration 2 or 3 times on my dev box without problems, so I'm pretty sure this can work on the production db as well. Ideally, one would like to copy the prod db locally and test the migration off-site, not sure how feasible that is, and in any case, I'd make full backups of everything so a rollback would be possible.

plural:
oh, for sure.  I'm going to do a local test to make sure i can safely deploy this and the MwlCard changes.  My caution is more about not wanting to knock NRDB offline and lack of experience with the schema updates than any lack of faith in your PR.  :D

We can definitely do the "decklists have multiple rotation stats" as a separate PR, though, so i'll approve this after i wrap up my local testing.  Thanks for the patch!!!

gereons:
I'll rebase this on main tomorrow to fix up the new rotation search

plural (approved):
i tested this out locally and it looks great.


The rotation search and rotation command are ... weird, but that is not your fault and we can deal with that later.  :D

PR #515, opened 2021-04-04
Split illustrators in the advanced search form.

plural:
this makes it easier to select all pieces by a single artist that collaborated with others on different cards.

Some data cleanup should happen in the card repo for a few odd ones, but that can happen separately from this.


NoahTheDuke (approved):
Should the splitting happen in a helper function some place so it can be called in the CardsData search function too?

plural:
ooh, i didn't realize how nicely wired up the services stuff is.  Moved the splitting logic to a new Illustrators service.

NoahTheDuke:
Ship it

PR #519, opened 2021-04-10
Do explicit === 0 check for banned cards to avoid type conversion.

plural:
silly type conversion.  my local env is on a slightly older rev of php 7.4 than the server.


NoahTheDuke (approved):
Hahaha

PR #525, opened 2021-04-15
Formats and mwl

plural:
Add Format and Ban List pages.  Re-organize the top nav menu a bit to make room.

The decklist links on the formats pages filter to the legal packs for each format. There is more we can do there, but this seems like a solid start.

Top Nav Menu
Screenshot 2021-04-14 9 44 31 PM

Formats Page
Screenshot 2021-04-14 9 45 15 PM

Banlist Page
Screenshot 2021-04-14 9 48 19 PM
Screenshot 2021-04-14 9 48 29 PM


app/Resources/views/Banlists/banlists.html.twig
```diff
@@ -0,0 +1,89 @@
...
+{% block body %}
+<div class="container">
+
+  <h1>{{ block('title') }}</h1>
+
+  <h2>Startup Format</h2>
+  <div id="list">
+    <div class="row">
+      <div class="col-sm-12">
+        <p>No cards are currently banned in Startup.  Have a blast!</p>
```
NoahTheDuke, line 18: Cheeky, i love it.

NoahTheDuke (approved):

PR #526, opened 2021-04-17
Show which cards are rotated in deckbuilder

cloudrac3r:
Instead of strugging to figure out which individual cards are rotated, now it just shows in the builder next to the card.

image

(Please forgive me for not having card images downloaded in my test environment.)

Notes:

- Is `🔄` or `🔁` the right symbol to use for rotated?
- This code should, in theory, show two icons if a card is rotated _and_ banned. In the picture you can see Security Leak (a Current) is not doing this for some reason. `MWL` in the browser console is `null`, so I think the code does actually work and it failed to load the banlist somehow? I'm still not familiar enough with the system to know what's happening here.

### DIFF
--- a/web/js/nrdb.js
+++ b/web/js/nrdb.js
@@ -258,20 +258,40 @@ function find_identity() {
 }
 
 /**
- * Returns a banned or restricted icon for the supplied card and selected MWL.
+ * Returns a banned, restricted, or rotated icon for the supplied card and selected MWL.
  * @param  card
  * @return string
  */
 function unicorn(card) {
     var mwlCard = get_mwl_modified_card(card);
 
+    var result = [];
+
+    function add_icon(icon, description) {
+        result.push('<span title="' + description + '" style="display:inline-block;width:1.5em;">' + icon + '</span>');
+    }
+
+    // Check MWL
     if (mwlCard.is_restricted) {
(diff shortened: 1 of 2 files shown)


cloudrac3r:
Here's a demo with all the icons working as intended.

image

web/js/nrdb.js
```diff
@@ -258,20 +258,40 @@ function find_identity() {
 }
 
 /**
- * Returns a banned or restricted icon for the supplied card and selected MWL.
+ * Returns a banned, restricted, or rotated icon for the supplied card and selected MWL.
  * @param  card
  * @return string
  */
 function unicorn(card) {
```
plural: how about updating this function name to something less colorful, like getCardIcons() or getCardLegalityIcons() ?  It looks like it's only directly referenced in this file.

plural:
heck yeah!

This looks great!  I just have one naming request.

Thanks a lot for contributing this!

web/js/nrdb.js
cloudrac3r: Sure thing!

plural (approved):

plural:
lovely, thank you!

I'll bundle this up with the next release (hopefully later this week).

PR #529, opened 2021-04-25
Fix PAD Factory Influence calculation, update a few other "latest printing" issues.

### LINKED ISSUE #528: PAD Factory influece not discounted with 3xPAD Campaign (System Update 2021) in deck
I suspect PAD Factory is checking for an old version of PAD Campaign while a deck is being built. It also does not surprise me that no one else has seen this error...

plural:
Fixes #528 

I switched some of this to use card names instead of codes, but a more thorough overhaul would be good, especially for the ext-decklist.js stuff.

### DIFF
--- a/web/js/nrdb.js
+++ b/web/js/nrdb.js
@@ -650,7 +650,7 @@ function get_influence_cost_of_card_in_deck(card) {
             }
         } else if (card.code === '10038') {
             // PAD Factory: 3 PAD Campaign
-            if (count_card_copies(NRDB.data.cards.find({ indeck: { '$gt': 0 }, code: ['01109', '20128','25142'] })) === 3) {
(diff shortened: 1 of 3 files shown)


web/js/nrdb.js
```diff
@@ -638,22 +638,22 @@ function get_influence_cost_of_card_in_deck(card) {
         if (Identity.code == "03029" && card.type_code == "program") {
             // The Professor: first program is free
             inf = (card.indeck - 1) * card.faction_cost;
-        } else if (card.code === '10018') {
+        } else if (card.code === 'Mumba Temple') {
```
francescopellegrini:
```suggestion
        } else if (card.title === 'Mumba Temple') {
```

Right?

plural: oops!  Yes, indeed.

I *swear* i tested this with temple as well.  Thanks for catching this!

PR #531, opened 2021-04-27
Update banlist 21.05 to have the right formatting for currents.

plural:


plural:
i need to add some metadata to the upstream JSON repo for this, but this will do for now.

Issue #308, opened 2019-01-27
Link to Errata, Rotation and MWL should probably point to NISEI.NET and not FFG


AndersRask:
Check for example this deck that uses a (NISEI) rotated Terminal Directive card:

https://netrunnerdb.com/en/decklist/53942/-612-welcome-to-the-doublemeat-palace

In the top there is a warning that say that

> This decklist is not legal for tournament play currently, due to Card Errata, Rotation or the Most Wanted List.

The links goes to FFG and not NISEI

NoahTheDuke:
Great idea! I’ll fix it!

plural:
![Screenshot 2021-06-27 6 44 12 PM](https://user-images.githubusercontent.com/396562/123563007-c4eeda80-d777-11eb-8b9d-914fe7f5aa7b.png)

One place where this shows up...

PR #532, opened 2021-05-14
replace each MWL reference with Legality

francescopellegrini:
There were a few places where the "MWL" term was still used.

### DIFF
--- a/app/Resources/views/Decklist/decklist.html.twig
+++ b/app/Resources/views/Decklist/decklist.html.twig
@@ -24,7 +24,7 @@ NRDB.user.params.decklist_id = Decklist.id;
 <div class="container decklist">
 
 {% if not decklist.is_legal %}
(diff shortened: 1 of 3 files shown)


francescopellegrini:
Also #308 could be worth a fix.

francescopellegrini:
I just noticed that both [deck search syntax](https://netrunnerdb.com/en/syntax) and [advanced search](https://netrunnerdb.com/en/search) refer to the terms "Banlist" / "Ban List" instead.. which one should we prefer?

plural:
we should probably use Ban List or ban list as two words consistently.

francescopellegrini:
> we should probably use Ban List or ban list as two words consistently.

@plural so would it help if I open a new PR to update all terms to "Ban List"? ;)

plural:
sure thing that would be swell!

-jason

PR #539, opened 2021-05-30
Return deck tags

North101:
It is possible to set tags from the api but they aren't returned.

This is a simple PR to do that.


plural:
thanks for the patch!

PR #541, opened 2021-05-31
Add agenda points icon to relevant agenda displays

plural:
Screenshot 2021-05-30 7 22 27 PM
Screenshot 2021-05-30 7 22 41 PM
Screenshot 2021-05-30 7 23 17 PM

### DIFF
--- a/app/Resources/views/Search/display-zoom.html.twig
+++ b/app/Resources/views/Search/display-zoom.html.twig
@@ -29,7 +29,7 @@
                     </span>
                     <span class="card-cost">
                         {% if card.type_code == "agenda" %}
-                            {{ card.advancementcost }}/{{ card.agendapoints }}
+							{{ card.advancementcost }}/{{ card.agendapoints }}<span class="icon icon-agenda-points"></span>
                         {% endif %}
                         {% if card.type_code != "identity" and card.type_code != "agenda" %}
                             {{ card.formatted_cost|raw }}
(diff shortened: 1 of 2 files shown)


app/Resources/views/Search/display-zoom.html.twig
```diff
@@ -29,7 +29,7 @@
                     </span>
                     <span class="card-cost">
                         {% if card.type_code == "agenda" %}
-                            {{ card.advancementcost }}/{{ card.agendapoints }}
+							{{ card.advancementcost }}/{{ card.agendapoints }}<span class="icon icon-agenda-points"></span>
```
NoahTheDuke: Looks like you used tabs instead of spaces

NoahTheDuke:
This is cool! Looks a little odd to have a single icon (when both halves are associated with icons). Should you add in the "cost" icon to the other side?

plural:
i don't have an advancement cost icon handy.  There isn't one in the visual assets set .  Using the credit icon isn't wrong, but it isn't right enough either.  :D

I'll ask on slack to see if i can get the asset and see how it looks.

app/Resources/views/Search/display-zoom.html.twig
plural: done, thanks.

plural:
i checked and we don't don't have an advancement icon to use.  I'm gonna go ahead and submit this.

Issue #520, opened 2021-04-12
Trying to run the project: Card import step: Call to a member function getCode() on null


cloudrac3r:
Hi, I'm trying to run the project so that I can actually test new features that I add. I'm familiar with PHP itself, but not any of the tools that surround its development like Composer and Symfony. I'm trying my best.

In the import step, when I have to run `php bin/console app:import:std -f /app/netrunner-cards-json` - the cards repository was cloned to the path `/app/netruner-cards-json` - I see this in the logs:

```
[truncated]
netrunnerdb_1  | Importing Cards...
netrunnerdb_1  | Changing the code of  
netrunnerdb_1  | Changing the deckLimit of 11001 
netrunnerdb_1  | Changing the position of 11001 
netrunnerdb_1  | Changing the quantity of 11001 
netrunnerdb_1  | Changing the strippedTitle of 11001 
netrunnerdb_1  | Changing the title of 11001 
netrunnerdb_1  | Changing the uniqueness of 11001 System Outage
netrunnerdb_1  | Changing the illustrator of 11001 System Outage
netrunnerdb_1  | Changing the keywords of 11001 System Outage
netrunnerdb_1  | Changing the text of 11001 System Outage
netrunnerdb_1  | Changing the strippedText of 11001 System Outage
netrunnerdb_1  | Changing the cost of 11001 System Outage
netrunnerdb_1  | 
netrunnerdb_1  | In ImportStdCommand.php line 510:
netrunnerdb_1  |                                                
netrunnerdb_1  |   Call to a member function getCode() on null  
netrunnerdb_1  |                                                
netrunnerdb_1  | 
netrunnerdb_1  | app:import:std [-f|--force] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command> <path>
```

When I try to use the web interface, none of the cards exist, which aligns with my assumption that the cards weren't imported correctly.

What can I do about this?

plural:
hey there!  Sorry you ran into trouble.

It looks like a fix to quiet down noisy messages on subsequent updates broke things for you here.

on 510, take out "&& ($entity->getType()->getCode() == 'identity' || $entity->getType()->getCode() == 'agenda')" and give it another whirl.

[...]

PR #550, opened 2021-07-21
Add setting to allow unlimited copies of cards

cloudrac3r:
Someone said they wanted this feature, and it seems like it would be good to add. https://www.reddit.com/r/Netrunner/comments/ooge0a/including_multiple_astroscripts_in_netrunnerdb/

Basically, when enabled in the settings tab, card text like "limit 1 per deck" no longer applies, and 3 copies (or more if the card text says more) can be added to the deck. A warning "too many copies of a limited card" is then shown. This warning was already in the code.

Questionable parts of this code:
- I have not tested card suggestions at all
- The for loop that generates the card count radio buttons is copypasted in 3 places. I'd like to merge them into 1, but not sure where I would put that function, since it needs to be shared across files, and I don't know how you're managing JS imports in netrunnerdb.

I think the rest of my code is all good, though. I tested both the card selection table and the card modal and they're good.

(By the way, [reverting this commit as discussed](https://github.com/NetrunnerDB/netrunnerdb/issues/520#issuecomment-821757580) is still required to create the database.)

### DIFF
--- a/app/Resources/views/Builder/deck.html.twig
+++ b/app/Resources/views/Builder/deck.html.twig
@@ -304,8 +304,8 @@ var Filters = {},
       </fieldset>
       <fieldset>
         <div class="radio"><label><input type="radio" name="card-limits" value="legal" data-persistence> Legal limit</label></div>
-        <div class="radio"><label><input type="radio" name="card-limits" value="ignore" data-persistence> Ignore card limits (casual play)</label></div>
-        <div class="radio"><label><input type="radio" name="card-limits" value="max" data-persistence> Maximum 9 (casual play)</label></div>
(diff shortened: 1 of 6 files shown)


distributive:
Sorry for the radio silence on this!

I've finally had a look at this, tested it, and cleaned it up a bit. Thanks for the contribution!

Changes:
- Options expanded to three (legal, ignore card text limits (e.g. limit 1 per deck), max 9)
- Draft IDs no longer force all cards to be l9pd
- Changed the layout of the settings menu to be a bit more concise and clear

cloudrac3r:
Yay!

plural (approved):
Thanks for the contribution!

PR #555, opened 2021-09-27
Remove ads.

### LINKED ISSUE #554: Remove Adverts
Ads are pointless, please yeet them into the sun

<3

plural:
Closes #554

### DIFF
--- a/app/Resources/views/Default/donators.html.twig
+++ b/app/Resources/views/Default/donators.html.twig
@@ -7,7 +7,7 @@
         <div class="col-md-12">
             <h1>The Gracious Donators</h1>
             <p class="lead">NetrunnerDB wouldn't be the same without them. Many, many thanks.</p>
-            <p>To show our appreciation, the content of the site is free of ads for Donators.</p>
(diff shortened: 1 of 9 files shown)


app/Resources/views/Default/donators.html.twig
```diff
@@ -7,7 +7,7 @@
         <div class="col-md-12">
             <h1>The Gracious Donators</h1>
             <p class="lead">NetrunnerDB wouldn't be the same without them. Many, many thanks.</p>
-            <p>To show our appreciation, the content of the site is free of ads for Donators.</p>
+            <p>To show our appreciation, the content of the site was free of ads for Donators, but as of September 2021, the site is permanently ad-free. We are working on new donator perks, so please stay tuned!</p>
```
NoahTheDuke: Maybe this should be something like "This website is ad-free due to the generous donations of our Patreon subscribers."

NoahTheDuke (approved):

app/Resources/views/Default/donators.html.twig
plural: yep.  That's great.

PR #561, opened 2021-10-30
Add caching to all public API methods except deck.

plural:
This takes care of 99% of the requests for the public APIs.

I should be able to change the qps quotas for these requests after this is deployed.  Right now this will use the default filesystem cache, but we should be able to configure it for something faster like memcache or even move the filesystem cache to a ramdisk if we want.

The Listener is needed to handle cache misses for decklists.  For everything else, we already clear the cache when importing new cards or data from the JSON repo.


NoahTheDuke (approved):
This is outside of my knowledge base but yolo!

PR #570, opened 2022-01-09
Add Deck & Decklist UUIDs

plural:
This is the first step in switching away from auto-incrementing IDs for decks and decklist.

This will prevent sequential scraping and (successfully) guessing at deck and decklist IDs.

We'll need followups for durable permalinks from this version of NRDB to v2.

### DIFF
--- a/src/AppBundle/Command/SetDecklistsUuidCommand.php
+++ b/src/AppBundle/Command/SetDecklistsUuidCommand.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace AppBundle\Command;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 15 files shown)


src/AppBundle/Command/SetDecklistsUuidCommand.php
```diff
@@ -0,0 +1,49 @@
...
+        $this
+                ->setName('app:uuid:decklists')
+                ->setDescription('Set a UUID for any decklist missing it.')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+		$sql = "UPDATE decklist "
+			. "SET uuid = LOWER(CONCAT("
```
NoahTheDuke: If you imported/brought in a uuid library, we should use it here. There's complicated math that goes into generating them that I don't think you're respecting in this implementation.

plural: hehe, not as much fancy math as you might think!

I fleshed out the command with an explanation and links to the spec.

PR #571, opened 2022-01-16
ID filter tools in new deck page

distributive:
Adds a set of options for specifying format and faction, as well as filtering by name, in the ID selection page for new decks

image

### DIFF
--- a/app/Resources/views/Builder/initbuild.html.twig
+++ b/app/Resources/views/Builder/initbuild.html.twig
@@ -2,28 +2,7 @@
 
 {% block body %}
 
-<script type="text/javascript">
-$(function() {
-    $('.list-group').on('mouseenter touchstart', 'a', function(event) {
-        var card_code = $(this).data('code');
-        var card = NRDB.data.cards.findById(card_code);
-        $('#cardimg').prop('src', card.imageUrl);
-    });
-    $(window).scroll(function() {
-        var scrollingDiv = $("#initIdentity");
-        if(!scrollingDiv.is(':visible')) return;
-        var y = $(this).scrollTop(),
-            maxY = $('footer').offset().top,
-            scrollHeight = scrollingDiv.height(),
-            scrollTop = $(window).scrollTop();
-
-        if(y < maxY - scrollHeight - 200) {
-            scrollingDiv.stop().animate({"marginTop": scrollTop + "px"}, "slow");
-        }
-    });
-});
-
-</script>
+<script src="{{ asset('/js/initbuild.js') }}"></script>
 
 <div class="container">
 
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Hey hey! Thank you for this! Mind posting some pictures of how it looks/works?

app/Resources/views/Builder/initbuild.html.twig
```diff
@@ -32,18 +32,55 @@ $(function() {
 <div class="col-sm-8">
 <div class="panel panel-default">
   <!-- Default panel contents -->
-  <div class="panel-heading"><b>Choose your identity</b> <span>(🔄  rotated card, 🚫  banned card)</span></div>
+  <div class="panel-heading"><b>Choose your identity</b> <span>(🔄  rotated card, 🚫  banned in standard)</span></div>
```
plural, line 35: nit: as we start to have proper multi-format support, it won't just be standard with bans on the site.

web/js/initbuild.js
```diff
@@ -0,0 +1,87 @@
...
+        var y = $(this).scrollTop(),
+            maxY = $('footer').offset().top,
+            scrollHeight = scrollingDiv.height(),
+            scrollTop = $(window).scrollTop();
+
+        if(y < maxY - scrollHeight - 200) {
+            scrollingDiv.stop().animate({"marginTop": scrollTop + "px"}, "slow");
+        }
+    });
+    function updateFormat(format) {
```
plural, line 34: A small improvement would be to clear out the ID image when the options change.
The ID images sort of linger, even when for example you had moused-over adam, then selected Startup for a format, with a similar effect when switching factions.
not required, but could be nice.

plural (approved):
I patched this in and it is a treat to use!

I have a couple of small suggestions - if you wanna take care of them now, let me know, otherwise, this looks great to me.  

:)

-jason

web/js/initbuild.js
distributive: This is a bit inelegant but I can make it hide the preview when the mouse stops hovering IDs, so you can't see a preview while changing settings

app/Resources/views/Builder/initbuild.html.twig
distributive: I'm not actually sure what the best way to show banned status is. There's an argument for not showing it since cards are filtered by legality anyway, but since the default view is all IDs it might be worth showing some indication.

web/js/initbuild.js
plural: it's up to you.  Let me know if you would like to take a stab at that here or as a followup.  I'm good either way.

distributive: I managed to get it to check if the ID has become invalid each time the list is updated, so if it all looks good now I should be done

plural: yesssssssss.

this looks fantastic!  Merging now.  :)

Issue #562, opened 2021-11-02
Change the way deck legality is displayed when viewing decks.


Tamijo-Netrunner:
![image](https://user-images.githubusercontent.com/39098450/139848590-082083a1-3682-44db-aa4b-e25abc4b9690.png)

Currently the deck legality section takes up a huge amount of screen space, more than a full screen on my smartphone. Much of this information is useless, and in some cases actively misleading. In this example above the deck is labelled as legal for MWL 1.0, which while technically true, isn't relevant. Most of these cards weren't released, so it was never actually legal under that list.

plural:
Yep, i think the right thing to do here is to show currently active and upcoming SBL (if one exists) but hide the rest behind something, like we do on the deckbuilder page now.

one note for posterity:

deckbuilding application of the banlist/MWL does not take into account cards that were released at a given time. It never has, but it wasn't noticeable back in the day.  NRDB has some unfortunate assumptions that make some of this all more confusing than it needs to be.  we as players have a notion of cards released at the time, but the deckbuilding/legality code never has.

PR #573, opened 2022-01-20
Make legality history collapsible in decklist menus

distributive:
Public [decklist pages](https://netrunnerdb.com/en/decklist/68657/lucky-lat-2nd-january-startup-tournament-#table-mwl) currently display the legality of their decklist over all banlists/MWLs/etc in NR history. This makes all but the most recent legality list hidden by default (or the two most recent if the most recent list isn't yet active).

Should resolve issue #562.

image

image


NoahTheDuke (approved):
This looks good to me.

plural (approved):
This is great.  Thanks!

PR #575, opened 2022-01-24
Full UUID support

plural:
This is a big pull request, but it was hard to break it up more usefully.  Sorry this is so large!

Update all the visible stuff to use UUID instead of ID for decks and decklists.  The thinking here is that it is fine to use ID on the backend, but we want the public identifiers to move to UUIDs.

Add permanent redirects (301s) for the deck and decklist ID urls to their URL equivalents.

Lastly. I moved the affected .js files to .v2.js files to break the cache when we push and prevent any errors.  This is annoying and hacky and manual, but it's easier than shoe-horning versioning in here as part of this pull request.


web/js/decks.v2.js
```diff
@@ -238,18 +238,16 @@ function do_action_deck(event) {
...
+  $('#decks a.deck-list-group-item.selected').each(function (index, elt) { uuids.push($(elt).data('uuid')); });
+  if(!action_id || !uuids.length) return;
   switch(action_id) {
-    case 'btn-compare': do_diff(ids); break;
-    case 'btn-compare-collection': do_diff_collection(ids); break;
-    case 'btn-tag-add': tag_add(ids); break;
-    case 'btn-tag-remove-one': tag_remove(ids); break;
-    case 'btn-tag-remove-all': tag_clear(ids); break;
-    case 'btn-delete-selected': confirm_delete_all(ids); break;
-    case 'btn-download-text': download_text_selection(ids); break;
```
NoahTheDuke, line 251: There are some removed lines here. Is that on purpose?

plural: yes!  Intentional!  

I did have to double-check, tho.  :D

There are individual deck download buttons, but no UI element for downloading all selected decks here so the download_text_selection and download_octgn_selection cases were dead.

Thanks for the careful review!

NoahTheDuke: Glad to help! I don't know enough to check the logic sadly, we'll have to trust you on that!

PR #577, opened 2022-01-24
Decklist selection

### LINKED ISSUE #565: "Compare decks" function seems to not work at all
In the my decks window the "with selection" dropdown opens but neither of the compare buttons seem to do anything.

### LINKED ISSUE #572: Feature Requests - Is it possible to make the compare decks function mobile friendly?
Currently on the My Decks page, in order to compare multiple decks you need to select them by holding shift which you can't do on a phone. Is there currently a mobile friendly way or could you add a mobile friendly way to select multiple decks, possibly a checkbox on each deck or something like that?

Thanks!

distributive:
- Adds a checkbox to decks in the private deck selection page
  - Makes the functionality clearer 
  - Allows decks to be selected on mobile
- Removed the ability to select decks by shift clicking
- Updated the help text and moved it to the top of the page
- Added select/deselect all buttons
- Fixed deck duplication not copying the tags of the original decklist

Should resolve #565 and resolve #572.

image

image


plural:
looks great!   Thanks!

PR #579, opened 2022-01-26
Remove unused claims controller and routes

plural:
i don't see any claims since 2018 and no traffic to this endpoint for as long as I can see either.


plural:
ok, gonna talk with necro about this.  It's apparently been broken for a while.  ;)

PR #587, opened 2022-02-07
Activity logs no longer show all user activity if the current user follows nobody

distributive:
Apparently the current workaround is to just follow a random user, which seems suboptimal.

This could be extended to show activity from comment threads you've interacted with.

### DIFF
--- a/src/AppBundle/Service/ActivityHelper.php
+++ b/src/AppBundle/Service/ActivityHelper.php
@@ -24,7 +24,7 @@ class ActivityHelper
         $following = $user->getFollowing();
 
         // don't return any items if the user follows nobody
-        if(!count($following)) {
+        if(count($following) == 0) {
             return [];
         }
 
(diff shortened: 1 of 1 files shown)


src/AppBundle/Service/ActivityHelper.php
```diff
@@ -21,9 +21,14 @@ public function __construct(EntityManagerInterface $entityManager)
 
     public function getItems(User $user, int $max_items = 30, int $nb_days = 7)
     {
-        $items = [];
-
         $following = $user->getFollowing();
+
+        // don't return any items if the user follows nobody
+        if(!count($following)) {
```
NoahTheDuke: I don't know php idioms well, but could this be count() == 0?

NoahTheDuke (approved):

src/AppBundle/Service/ActivityHelper.php
distributive: The original code had `if(count($following))` so I just copied the style. I don't know if there's a preferred way of doing this.

NoahTheDuke: Haha that's true. I prefer explicit over implicit but I won't block the PR over it

PR #589, opened 2022-02-07
Remove NAPD Multiplayer from the list of standard-legal cycles

distributive:
See: https://netrunnerdb.com/en/formats


NoahTheDuke (approved):

plural:
Thank you!

Issue #377, opened 2019-11-30
OAuth and multiple accounts


Absotively:
A user has reported that while logged into their personal account, they logged in to Always Be Running through NRDB, and found themselves logged into ABR as an organizational account they use sometimes.

plural:
do you have more details?  An organizational account for ... what exactly?

I am starting to look into some more oauth stuff, so if there is any more detail, I'm happy to take a look.

Absotively:
The organizational account was for NISEI. They had logged into ABR with NISEI’s account previously to set up some tournaments as NISEI.

plural:
oh, interesting.  Thanks for the extra details.  If you want to share any other details on slack we can see if we can reproduce it and debug.

plural:
Ok, #590 should take care of this and will be in the imminent release!

PR #590, opened 2022-02-08
OAuth Login Flow Improvements (show username, switch accounts)

### LINKED ISSUE #421: OAuth Screen should show what account you are logged in as
Several people have had trouble mixing up their accounts for things like ABR.  The auth flow doesn't show you which account you are logged in as.

It should, though!  That would be swell.

plural:
Update the OAuth flow to show the username for the logged in account and support a logout link and allow user switching.

This required updating the logout handler to clear sessions on logout.  Previously, logging out would clear the main application's session, but *leave the oauth session alone*.  Because the OAuth page didn't show the username, people with multiple OAuth logins could be surprised about which account was in use.

Screenshot 2022-02-07 10 41 54 PM

Should fix #421 and #377


NoahTheDuke:
Looks like a lot of overlap with #591

plural:
discussed / approved with noah over chat.  :)

PR #593, opened 2022-02-09
Search Improvements

### LINKED ISSUE #552: Draft cards show up in a standard search
https://netrunnerdb.com/find/?q=b%3Astandard-ban-list-21-06+z%3Arotation-2021+c%3A0&sort=name&view=list&_locale=en

The above link should show a blank search. Draft IDs are counted as "in rotation". They should be excluded from all standard format rotations.

### LINKED ISSUE #553: Feature Request: mini faction quick filter
Is there a way we could get "AD"/"AP"/"SL" added as filters for the three mini factions?

distributive:
- Resolves #552 - draft cards can no longer appear when a rotation is specified (this isn't a perfect solution but is unlikely to ever be a problem - the ideal fix would be to have draft and standard be explicit formats)
- Fixed `f:-` only finding neutral corp cards (now finds neutral runner cards as well)
- Added the following shortcuts: 
  - `nc` - Neutral corp cards
  - `nr` - Neutral runner cards
  - `neutral` - Alternative shortcut for all neutral cards
  - `ap`, `ad`, `su` - Alternative shortcuts for the mini-factions (resolves #553)
  - `mini` - All mini-factions
  - `hb` - HB
  - `weyland` - Weyland
- Updated and fixed the search syntax reference
- The search syntax reference is now linked when a search fails to find any cards

### DIFF
--- a/app/Resources/views/Default/syntax.html.twig
+++ b/app/Resources/views/Default/syntax.html.twig
@@ -52,13 +52,14 @@
         </li>
         <li><b>Faction</b> <code>f</code> accepts full faction codes (e.g. nbn, haas-bioroid, weyland-consortium, sunny-lebeau, etc) or the following shortcuts:
           <ul>
-            <li>The first letter of each non-neutral faction (or the second letter for mini-factions)</li>
+            <li>The first letter of each non-neutral, non-mini factions</li>
+            <li>The second letter, or first two letters, of mini-factions (e.g. <code>p</code> or <code>ap</code> for Apex)
+            <li><code>mini</code> for all mini-factions</li>
             <li><code>nc</code> for <code>neutral-corp</code></li>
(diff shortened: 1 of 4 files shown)


NoahTheDuke (approved):
Looks good to me. Can't believe I got the hash array backwards and overwrote neutral runner.

NoahTheDuke:
Oh, make sure to update the tooltip syntax reference too. It's slightly different which is annoying.

distributive:
> Oh, make sure to update the tooltip syntax reference too. It's slightly different which is annoying.

What's the best way to fit it concisely in? Does this work:
`f - faction (faction code, letter, or '-' for neutral)`

NoahTheDuke:
Yeah that seems short and sweet to me. I remember making concessions like that for the others (on my phone or I'd look it up lol).

plural (approved):
This is great!

PR #594, opened 2022-02-10
Search improvements

### LINKED ISSUE #342: Nickname of Cards
It would be nice to have the "official" nick names of cards listed here. So they could appear in search and for new players to see and learn them.
Like "Kitty" for QPM or "Baby" for "Symmetrical Visage"
Sure this is an enhancement that might not have priority.

distributive:
- Added fallbacks for searches that fail to find any cards when no criteria (`x:...`, `o:...`, etc) has been set:
  - After the first fail, it will treat the input as an acronym and try again
  - If that fails, it will try to match the input as the substring of a card alias
- Added a list of card aliases: [full list](https://github.com/NetrunnerDB/netrunnerdb/blob/62edf0de65130a460caad44ed478b840971f3594/web/card_aliases.txt)
- Documented this feature in the search syntax reference page, as well as the ability to search for card acronyms
- Added `_` as an explicit criteria for searching titles (including the ability to search for titles that don't match a query)
- `z` now accepts "startup" as a parameter, and supports the `!` operator

Resolves #342 

New alias section on syntax reference page:
image


web/card_aliases.txt
```diff
@@ -0,0 +1,129 @@
...
+tfp : The Future Perfect
+mca : MCA Austerity Policy
+mcaap : MCA Austerity Policy
+mca ap : MCA Austerity Policy
+emcap : MCA Austerity Policy
+wtwb : Watch the World Burn
+baby : Symmetrical Visage
+butler : Jeeves Model Bioroids
+pancakes : Adjusted Chronotype
+cell phones : 02069
```
NoahTheDuke, line 98: Any reason for card id here instead of card title?

web/card_aliases.txt
```diff
@@ -0,0 +1,129 @@
+franklin : Crick
+tfp : The Future Perfect
+mca : MCA Austerity Policy
+mcaap : MCA Austerity Policy
+mca ap : MCA Austerity Policy
```
NoahTheDuke, line 20: How does this work if the search is `mca x:a ap`? Does the entire search have to be exactly `mca ap`? Allowing access makes me think we could open ourselves to subtle bugs or functionality we don't intend.

web/card_aliases.txt
```diff
@@ -0,0 +1,129 @@
...
+slimhack : Overclock
+desperadito : Pennyshaver
+fao : Forged Activation Orders
+bbg : Breaker Bay Grid
+blorch : Black Orchestra
+blorc : Black Orchestra
+tpr : Timely Public Release
+sva : Salvaged Vanadis Armory
+midseasons : Midseason Replacements
+sft : Shipment from Tennin
```
NoahTheDuke: Should the whole file be alphabetical? Might make adding/removing aliases easier in the future.

app/Resources/views/Default/syntax.html.twig
```diff
@@ -50,6 +50,13 @@
             <li><code>&gt;</code> &ndash; more than</li>
           </ul>
         </li>
+        <li><b>Title</b> - If you enter text with no criteria, it will attempt to match against card titles (e.g. <code>mca</code> will match MCA Austerity Policy and MCA Informant). The only exceptions are:
+          <ul>
+            <li>Two-or more letters in all capitals will attempt to match cards by acronym (e.g. <code>BLC</code> will match Blue Level Clearance and Black Level Clearance)</li>
+            <li>Writing the ID number of a card will match that card specifically (e.g. <code>20073</code> will match Green Level Clearance)</li>
+            <li>Some cards have aliases that will match them (e.g. <code>franklin</code> will match Crick)
```
NoahTheDuke: Maybe the example should be something that's in use in the community already: clippy for Paperclip.

distributive: Good idea - I just picked these examples arbitrarily

web/card_aliases.txt
distributive: The phones aspect refers to specifically the original printing

web/card_aliases.txt
distributive: Sure, although I'd probably split that into two with acronyms separately - ideally they can be made automatic at some point

web/card_aliases.txt
distributive: It doesn't work, which is actually weird because I think it should based on the implementation. Either way it's not too important since I've filtered out any alias that would otherwise get you real results (I'll make it automate this so we don't have to do this manually every time cards get released)

NoahTheDuke:
Damn, this looks good as hell.

src/AppBundle/Service/CardsData.php
```diff
@@ -1117,4 +1146,31 @@ public function get_versions_by_code(array $cards_code)
...
+
+    public function getPrettyCardAliases()
+    {
+        // Construct mapping of cards to their aliases
+        $cardAliases = [];
+        foreach ($this->cardAliases as $alias => $card) {
+            // Translate card codes into their card's name
+            $code = preg_match('/^\d\d\d\d\d$/u', $card);
+            if ($code) {
+                $card = $this->entityManager->getRepository(Card::class)->findBy(['code' => [$card]])[0]->getTitle();
```
NoahTheDuke: Query in a loop is worrisome.

Maybe added a `cardCodes = [];` outside the loop then push into it when finding codes instead of pushing the title, and then do a single query outside of the loop and loop over the results to do the same as the original loop?

I realize it's more annoying to write it this way but it stops potentially performing _lots_ of queries when loading this page.

src/AppBundle/Service/CardsData.php
```diff
@@ -650,25 +668,41 @@ public function get_search_rows(array $conditions, string $sortorder, string $lo
...
-                        $rotation = $rotationservice->findCurrentRotation();
-                    } elseif ($condition[0] == "latest") {
-                        $rotation = $rotationservice->findLatestRotation();
-                    } else {
-                        $rotation = $rotationservice->findRotationByCode($condition[0]);
-                    }
-                    if ($rotation) {
-                        // Add the valid cycles for the requested rotation and add them to the WHERE clause for the query.
-                        $cycles = $rotation->normalize()["rotated"];
+                    if ($condition[0] == "startup") {
```
NoahTheDuke, line 671: I wonder if this should be moved into RotationService, like findStartupRotation or something. That way we're keeping all of the rotation-finding logic together instead of some here and some in RotationService.

NoahTheDuke (approved):
Looks great tbh. Gonna greatly improve the usability of the basic search.

src/AppBundle/Service/CardsData.php
distributive: I couldn't work out how to make it work in RotationService. There's [another function](https://github.com/NetrunnerDB/netrunnerdb/blob/8cf43d59790c20b4610b3c8162534d2fb9e4dd31/src/AppBundle/Controller/SearchController.php#L497) that defines startup which also does it by manually defining it on top of the RotationService stuff.

PR #600, opened 2022-02-24
Add deck_uuid param to saveDeckAction and publishDeckAction

### LINKED ISSUE #599: Option to save deck by uuid

North101:
Fixes #599

Note: I haven't tested this because I am not sure how to run it locally.

### DIFF
--- a/src/AppBundle/Controller/PrivateApi20Controller.php
+++ b/src/AppBundle/Controller/PrivateApi20Controller.php
@@ -169,6 +169,7 @@ class PrivateApi20Controller extends FOSRestController
      *  description="Save one (private) deck of authenticated user",
      *  parameters={
      *    {"name"="deck_id", "dataType"="integer", "required"=true, "description"="ID"},
+     *    {"name"="deck_uuid", "dataType"="string", "required"=true, "description"="UUID"},
      *    {"name"="name", "dataType"="string", "required"=true, "description"="Name"},
      *    {"name"="description", "dataType"="string", "required"=false, "description"="Description of the deck in Markdown"},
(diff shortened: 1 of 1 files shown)


NoahTheDuke (approved):

NoahTheDuke:
Should we be worried about it choosing uuid if both are provided?

North101:
> Should we be worried about it choosing uuid if both are provided?

So the reason I wrote it that way is because I was under the assumption that you were moving from id to uuid. Im happy to flip the logic if you would prefer the other way around.

NoahTheDuke:
Ah yeah, that's a good point. Cool, then I think this is good.

plural:
let me patch this in and take it for a spin tonight.

plural:
I'm still doing some testing (and getting distracted by trying to work up some instructions for local oauth testing...), but there is a syntax issue to fix.

North101:
> I'm still doing some testing (and getting distracted by trying to work up some instructions for local oauth testing...), but there is a syntax issue to fix.

Woops fixed.

North101:
> Should we be worried about it choosing uuid if both are provided?

So one option is to error if both are provided

plural:
Hiya!  Sorry for the radio silence on this.  I did end up patching both of your PRs in and making a couple changes.  I need to grab the machine i did that on and get you and update. 

Largely this is good, but it exposed a couple rough edges for us.

Thanks for the patience *and* for the contributions!

locks:
@plural anything I can help with?

plural:
I apologize for letting this linger as long as it has.  With the progress on the new versions of NRDB (v2) and the API (v3), i'm going to close this one out.  That new API is UUID only to start.

PR #607, opened 2022-03-07
Add Borealis set icon.

plural:
Screen Shot 2022-03-06 at 8 06 37 PM
Screen Shot 2022-03-06 at 8 06 44 PM
Screen Shot 2022-03-06 at 8 06 54 PM


web/netrunnerfont.css
```diff
@@ -135,6 +135,9 @@
 .icon-system-update-2021:before {
   content: "\e92e";
 }
+.icon-borealis:before {
+  content: "\e925";
```
NoahTheDuke: This is already taken by kitara, so that's why the svg has such a big diff.

NoahTheDuke (changes requested):
Should the composer.lock file be updated here? I don't see any reason for the change.

web/netrunnerfont.css
plural: nope, the composer.lock change was just an accident.  Reverted.

Re: the SVG.  grrrrrrrrr.  the process for doing this is janky and apparently i did it wrong.  I'll look more closely tonight.

NoahTheDuke: yeah it's really really really annoying lol. i think it's append-only? can't insert in the middle of the list.

plural: ah, i figured out what happened.  I had imported the icon twice after fiddling with the size.  I left one accidentally in the middle of the grid.  Ooops!  Turns out if you do that and remove it, icomoon doesn't reset the number properly for you.  Live and learn.

NoahTheDuke (approved):

PR #609, opened 2022-03-09
Use old function for the old php version on the server.

plural:


NoahTheDuke:
Can we update the php version lol? Or will that break everything

plural:
it won't break everything, but it is more invasive.

plural:
we can definitely upgrade to php 8 if we really, really wanted to, but i definitely don't really, really want to.  :D

Issue #602, opened 2022-02-28
Slightly unclear wording in syntax guide.


Tamijo-Netrunner:
![unknown-2](https://user-images.githubusercontent.com/39098450/155963089-0254a7d5-f949-4c98-b149-b4839aa13c9d.png)

The above section could use OR instead of AND to be clearer. (There may be more examples on the page that have been missed, too.)

PR #610, opened 2022-03-09
Minor search fixes

distributive:
- Removed some card aliases that interact weirdly with the search syntax
- Documented `z:startup` as a valid search query
- Clarifies some wording in the syntax guide (resolving #602)


NoahTheDuke:
Bummer about mr phones. I wonder why that's happening

PR #612, opened 2022-03-19
More minor search fixes

distributive:
- Fixed negative conjunctions in title queries not working (e.g. `_!a|e|i|o|u`)
- Fixed some queries being matched as acronyms when they shouldn't be (e.g. `mr phones` matched cards with the acronym **MR**)


NoahTheDuke:
I think this is right but you're far past what I know about search.

PR #619, opened 2022-04-18
Update card image code to pull from a configured URL.

plural:
this is prep for switching to the CDN for card images.

There will be a bit more work to switch formats (jpg -> png) or to switch any sizes.


NoahTheDuke:
To double check, `NRDB.card_image_url` is already defined someplace in the javascript?

plural:
It is now in the layout twigfile.

NoahTheDuke (approved):

NoahTheDuke:
Ope, good catch.

PR #624, opened 2022-04-25
Updated the way card legalities are displayed

distributive:
- Removed emojis and replaced them with colour-coded labels.
- Removed all references to the April Fools Roseville set

Some people find the use of emojis confusing, especially the 🔄 symbol for rotated cards. This is especially true of new players who might not understand what rotation means, and non-native speakers who may not make the connection between rotation and the arrows.

### Examples of the old design
With card
Banned card

### Examples of my proposed new design
With card
Available and unbanned in Startup and Standard
Not in Startup; banned in Standard
image

### Deckbuilder
image
#### On mouseover
image

### ID selection
image

### Faction pages
image


app/Resources/views/Search/display-zoom.html.twig
```diff
@@ -125,19 +125,40 @@
                             <th class="legality-{{ card.startup_legality }}"> Startup Card Pool</th>
                         </tr>
                         <tr>
-                            <th class="legality-{{ card.standard_legality }}"> Standard Card Pool</th>
+                            <th
+                                {% for mwl_info in card.mwl_info %}
+                                  {% if mwl_info.active %}
+                                    {% if card.standard_legality == 'available' %}
+                                      class="legality-{{ mwl_info.legality }}"
```
NoahTheDuke, line 132: Is there a way we could move this into a php function that we call in the preparation, which we could reference here? Then it's just `<th class="legality-{{ card.legality_class_string }}">`. That way, we leave as much logic outside of the twigs as possible.

NoahTheDuke (approved):
I prefer to keep logic out of twig files but I'm not really working on this anymore so I'll leave it up to your discretion.

plural:
Overall this looks great!  Thanks!

Restricted looks quite tight, but the colors and the text look good.

I'm not sure the note about not being in rotation helps, though, tbh. The interactions between banlist and rotation are annoying right now, but a banlist and rotation are two independent things until one sets a deck{,list} specifies both of them.

We will also need to update some stuff in the deck builder, the ID selection pages and the faction if pages as well.

distributive:
> I'm not sure the note about not being in rotation helps, though, tbh. The interactions between banlist and rotation are annoying right now, but a banlist and rotation are two independent things until one sets a deck{,list} specifies both of them.

Because of the current limitations I made the standard ban list label say "NOT LEGAL" instead of "LEGAL" for rotated cards, but on the full history of the card it's still marked as "legal".

Without the note it looks a bit confusing:

image

app/Resources/views/Search/display-zoom.html.twig
distributive: I probably could, but I'd need to change some of the logic, and since we'll be moving to Ruby eventually anyway I'm less concerned with keeping this code optimal.

plural:
Yeah, I’m not a fan of not legal for the ban list, personally.  We don’t do careful bookkeeping for when a given banlist/rotation combo was in effect, so I am not sure this adds a lot of info.  Wil all banlist so show the right info about rotation or just the latest?

maybe others disagree, but unless a card is on a banlist, I still think the most correct thing is to say legal for that banlist.

maybe b2 will help us deal with this more elegantly.

plural:
2^W3 more pieces of feedback/requests:
- can the legend on traction page be on a single line?
- Can the faction label on the id line be closer to vertically centered or a bit bigger to help with that alignment?
- the labels on the deck builder card list are quite large in comparison and make it hard to read the list.
- 
thank you for adding the extra coverage here!

distributive:
I've made these changes. To fix them clogging up the decklist interface I've made them into icons that expand on mouseover (see above).

plural:
This is rad. One last thing I swear! How do the little icons work on mobile? Can you tap them to expand?

Thanks again. This looks swell!!!

distributive:
You can't, but the messages at the top of the decklist still appear:
image

EDIT: it now works on mobile

PR #626, opened 2022-04-28
Added an update log to the home page

distributive:
Adds a list of site updates to the home page:
image
image
This mirrors the update log jinteki.net has on its homepage.

In this implementation updates have to be added manually to `web/update_log.txt`. A database table would be cleaner, but ultimately requires the same steps but with additional tedium. If `web/update_log.txt` doesn't exist, the update log isn't shown at all.

### DIFF
--- a/src/AppBundle/Controller/IndexController.php
+++ b/src/AppBundle/Controller/IndexController.php
@@ -36,6 +36,28 @@ class IndexController extends Controller
         // recent decklists
         $decklists_recent = $decklistManager->recent(0, 10, false)['decklists'];
 
+        // load site updates
+        $file = fopen('update_log.txt', 'r');
+        $updates = [];
+        $update = null;
+        if ($file) {
+            while (($line = fgets($file)) !== false) {
+                $line = trim($line);
+                if (empty($line))
+                    continue;
+                if ($line[0] !== '-') {
+                    if ($update != null) {
+                        $updates[] = $update;
+                    }
+                    $update = array('date' => $line, 'entries' => []);
+                } else if ($update !== null) {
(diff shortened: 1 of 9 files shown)


src/AppBundle/Controller/IndexController.php
```diff
@@ -36,6 +36,30 @@ public function indexAction(Request $request, EntityManagerInterface $entityMana
         // recent decklists
         $decklists_recent = $decklistManager->recent(0, 10, false)['decklists'];
 
+        // load site updates
+        $updates = [];
+        if (file_exists('update_log.txt')) {
+            $file = fopen('update_log.txt', 'r');
+            $update = null;
```
NoahTheDuke: Should this be cached in some way? Maybe store `filemtime('update_log.txt')` as a class variable? Here's off-the-cuff code:

```php
class IndexController extends Controller
{
  var $update_log_timestamp;
  var $updates;
  ...

  if (is_null($this->updates)) {
    $this->updates = [];
  }

  $last_modified = null;
  if (file_exists('update_log.txt')) {
    $last_modified = filemtime('update_log.txt');
  }

  if (!is_null($last_modified) && $last_modified != $this->update_log_timestamp) {
    $file = fopen('update_log.txt', 'r');
    ...rest of your function...
  }
```

and then use $this->updates in your "read file" code. That way, once you've read it once, you won't read and process the file again, it'll just hang out as a small bit of memory in the app.

distributive: Good shout; that's now implemented

NoahTheDuke (approved):
:chef_kiss:

PR #631, opened 2022-05-10
Added set codes to the syntax page and also a close button to the update log and added `z:standard` as a valid search query as well

### LINKED ISSUE #620: Feature update: `z:standard`
Can we add "standard" to the list of applicable search options for `z:`? (And it would simply match 'current'.)

distributive:
A set code for every pack. A close button in every update log.

Resolves #620

image

### DIFF
--- a/src/AppBundle/Controller/DefaultController.php
+++ b/src/AppBundle/Controller/DefaultController.php
@@ -152,12 +152,14 @@ class DefaultController extends Controller
 
         $banlists = $entityManager->getRepository(Mwl::class)->findBy([], ['dateStart' => 'DESC']);
         $rotations = $entityManager->getRepository(Rotation::class)->findBy([], ['dateStart' => 'DESC']);
+        $packs = $entityManager->getRepository('AppBundle:Pack')->findBy([], ['dateRelease' => 'DESC']);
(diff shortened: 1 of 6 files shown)


src/AppBundle/Controller/DefaultController.php
```diff
@@ -152,12 +152,14 @@ public function syntaxAction(EntityManagerInterface $entityManager, CardsData $c
 
         $banlists = $entityManager->getRepository(Mwl::class)->findBy([], ['dateStart' => 'DESC']);
         $rotations = $entityManager->getRepository(Rotation::class)->findBy([], ['dateStart' => 'DESC']);
+        $packs = $entityManager->getRepository('AppBundle:Pack')->findBy([], ['dateRelease' => 'DESC']);
```
NoahTheDuke, line 155: Might we want to sort by name? Seems like it would be more helpful when searching for a specific code to scan alphabetically instead of having to consider _when_ it came out.

NoahTheDuke (approved):

src/AppBundle/Controller/DefaultController.php
distributive: I find them more confusing not in release order, but I have no strong opinion either way.

Issue #630, opened 2022-05-03
Feature request: Add the ability to remove the update box.


Tamijo-Netrunner:
Could we add some kind of cross to close the updates box, and then save that it has been closed via somebody's profile? It could (should?) then only reappear and need to be closed again when a new update is posted.

plural:
update log was moved below decklist of the week, so i am gonna close this request out.

PR #633, opened 2022-05-14
Move update log to the bottom of the homepage, above the footer.

plural:
I like the updates, but they won't be relevant enough to justify being at the top of the homepage, IMO.  Happy to do something else if we want.

Screen Shot 2022-05-14 at 12 27 41 PM

Mostly addresses #630


distributive (approved):
Apologies; I forgot to approve this sooner

plural:
no worries at all.  Thanks!

PR #637, opened 2022-06-01
Fixed some differences between the search syntax page and the deckbuilder search

distributive:
- Added MU cost and illustrator as valid search parameters in deckbuilding
- Documented `y` (quantity) in the syntax help page
- Reordered queries in the js search function to match the search syntax page
- Added a query for base link in runner IDs (e.g. `l:2` gets only Sunny)

### DIFF
--- a/web/js/nrdb.smart_filter.js
+++ b/web/js/nrdb.smart_filter.js
@@ -51,6 +51,9 @@
       case "v":
         add_integer_sf('agenda_points', operator, values);
         break;
+      case "l":
+        add_integer_sf("base_link", operator, values);
+        break;
       case "n":
         add_integer_sf('faction_cost', operator, values);
         break;
--- a/web/js/nrdb.smart_filter.js
+++ b/web/js/nrdb.smart_filter.js
@@ -20,18 +20,6 @@
       var values = condition;
 
       switch (type) {
-      case "e":
-        add_string_sf('pack_code', operator, values);
-        break;
-/*      case "c":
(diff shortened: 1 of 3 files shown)


web/js/nrdb.smart_filter.js
```diff
@@ -42,33 +30,54 @@
...
+      case "e":
+        add_string_sf('pack_code', operator, values);
+        break;
+      // case "c":
+      //   add_integer_sf('cyclenumber', operator, values);
+      //   break;
+      case "t":
+        add_string_sf('type_code', operator, values);
+        break;
+      // case "f":
```
plural: any reason to leave these commented?

plural (approved):

web/js/nrdb.smart_filter.js
distributive: The remaining parameters all already implemented through the UI. We could put implement them as an alternative way to search for cards, but we'd need to put in some safeguards because things can get weird (e.g. you can type `f:shaper` into a corp deck and switch to a shaper ID).

plural: Ah, thanks for the explanation!

Would you please add a comment to that effect instead of leaving the code commented for those cases?

It would be good, but not necessary, to update the syntax page as well.

distributive: How's this?
image

NoahTheDuke:
This has been needed for 5+ years. You're a goddamn champ for handling this.

plural (approved):
rad.  thanks!

PR #640, opened 2022-06-07
Fix n search op (deckbuilder was fine) to deal with 0 influence cards.

plural:
Thanks to @distributive for reporting this and the folks @ GLC for finding it.

We weren't handling this properly on the server-side search, but the deckbuilder was OK.
Before:

Screen Shot 2022-06-06 at 10 56 06 PM

Screen Shot 2022-06-06 at 10 56 46 PM

After:

Screen Shot 2022-06-06 at 10 52 12 PM

Screen Shot 2022-06-06 at 10 52 24 PM


distributive (approved):
Oh I didn't find it; I just reported it.

PR #642, opened 2022-06-08
Fixed deckbuilding search - queries are no longer parsed as regexes

### LINKED ISSUE #639: Syntax Issue in Deckbuilder Search Bar
This search works fine in the regular search outside the Deckbuilder page: x!"+1" s:icebreaker <- searching for all icebreakers not having a "+1" in the text

Anyway, the same search does not seem to work inside the Deckbuilder search bar and the part x!"+1" is automatically removed as an incorrect syntax.

distributive:
Resolves #639 
Also fixes negative searches not working on string parameters in the deckbuilder

### DIFF
--- a/web/js/nrdb.smart_filter.js
+++ b/web/js/nrdb.smart_filter.js
@@ -17,7 +17,9 @@
       var condition = conditions[i];
       var type = condition.shift();
       var operator = condition.shift();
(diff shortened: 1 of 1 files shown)


plural:
Great catch!

plural (approved):

PR #649, opened 2022-06-12
Tidy up docker config to allow running as prod as well.

plural:
this resolves an issue with accessing app.php instead of app_dev.php, which we need if we want to deploy prod via Docker.


plural:
This also removes the no-longer-used slack log webhook.

locks (approved):

PR #657, opened 2022-07-02
Added a border radius to card images in card pages

distributive:
This adds a border radius to card images on card pages, removing the non-transparent corners.

Examples:
image
image


plural:
what's it look like on some old and some new cards?

distributive:
I've added examples now.

Now I think about it, if we have convenient access to card art without black borders then updating our art is probably the more elegant solution, but if not then this will do that.

NoahTheDuke:
Damn, that looks great

plural:
nice!  

We have scans we can tidy up, but this is a nice bit of consistency for the site as well.

plural (approved):

PR #661, opened 2022-07-14
Brain/core interchangeability

distributive:
This makes it so that searching for "brain damage" or "core damage" in a card's text will get results for both terms. For versatility it specifically looks for "brain d" or "core d", but not "brain" or "core" alone as those would get false matches (e.g. agendas with on-score abilities).

### DIFF
--- a/src/AppBundle/Service/CardsData.php
+++ b/src/AppBundle/Service/CardsData.php
@@ -290,7 +290,6 @@ class CardsData
                     $or = [];
                     // Apply text aliases
                     foreach ($textAliases as $alias => $replacement) {
-                        print ($alias . " " . $replacement . " - ");
                         foreach ($condition as $arg) {
                             if (str_contains($arg, $alias)) {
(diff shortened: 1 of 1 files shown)


src/AppBundle/Service/CardsData.php
```diff
@@ -281,6 +288,16 @@ public function get_search_rows(array $conditions, string $sortorder, string $lo
                     break;
                 case 'x': // text
                     $or = [];
+                    // Apply text aliases
+                    foreach ($textAliases as $alias => $replacement) {
+                        print ($alias . " " . $replacement . " - ");
```
NoahTheDuke: 👀

plural: if this is removed, this works great.  :)

plural (changes requested):

src/AppBundle/Service/CardsData.php
distributive: Oops!

plural (approved):

PR #663, opened 2022-07-15
Adding set icon/numbers to cards in deck comparison

andreshomeroloz:
Hello there! First time contributor here. Got into Netrunner recently and just starting to get familiar with NetrunnerDB (and GitHub itself :P), so please bear with me.

When playing around with decklists I found the 'Compare' functions in the My Decks page. I thought it was pretty cool and useful to quickly see what needs to be replaced in order to turn a physical deck into a different version of it, except that for a new player like me it's kinda hard to know where I might be able to find a particular card without knowing what its set and number is. I saw in some other parts of the site that this info is sometime displayed on card lists when they're being sorted by set, so I thought I'd propose for them to be added here.

Aside from adding the icons and set numbers, I refactored the generation of these items into its own function in that same file, hopefully making any future changes here a little bit easier and more uniform.

I was not able to set up a proper local copy of the site, but I WAS able to run the modified JS code on the prod site directly through the Chrome dev console and these were the results:
Screen Shot 2022-07-14 at 11 46 07 PM
Screen Shot 2022-07-14 at 11 45 39 PM
Screen Shot 2022-07-14 at 11 45 25 PM
Screen Shot 2022-07-14 at 11 42 03 PM

I'd totally understand if this is something that might need more of a justification or more testing, but I thought it'd be cool to have.

### DIFF
--- a/web/js/decks.v2.js
+++ b/web/js/decks.v2.js
@@ -128,7 +128,7 @@ function do_diff(uuids) {
   var item_data = $.map(intersect, function(qty, card_code) {
     var card = NRDB.data.cards.findById(card_code);
     if(card) return { card: card, qty: qty };
-  }).sort(function (a, b) { return a.title.localeCompare(b.title); });
+  }).sort(function (a, b) { return a.card.title.localeCompare(b.card.title); });
   $.each(item_data, function (index, item) {
     list.append(get_card_list_item_html(item.card, item.qty));
   });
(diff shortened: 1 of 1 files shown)


distributive (approved):
Hi! Thanks for the contribution, this looks great! As far as I can tell this works fine. If you wanted to expand on this it might be nice if the cards were sorted by set rather than alphabetically.

If you're having trouble running a local dev version I'd suggest looking at this, it might be a little more manageable: https://github.com/Absotively/nrdb-dev-env

web/js/decks.v2.js
```diff
@@ -97,6 +97,10 @@ function decks_upload_all() {
   $('#archiveModal').modal('show');
 }
 
+function get_card_list_item_html(card, quantity) {
+  return '<li>'+quantity+'x '+card.title+' (<span class="small icon icon-' + card.pack.cycle.code + '"></span> ' + card.position + ')</li>';
```
NoahTheDuke:
```suggestion
  return '<li>' + quantity + 'x ' + card.title + ' (<span class="small icon icon-' + card.pack.cycle.code + '"></span> ' + card.position + ')</li>';
```
Extra spacing helps readability.

andreshomeroloz:
@distributive thanks! I do want to expand on this (maybe with a "sort by" selector somewhere), but wanted to get my feet wet first.

PR #685, opened 2022-09-04
New api format page

plural:
this has a ton of stuff hard-coded, but it works locally (assuming you are running the API server locally!).

I'll come back to make the config variables and whatnot, but any early feedback would be swell!

### DIFF
--- a/app/Resources/views/Formats/formats.html.twig
+++ b/app/Resources/views/Formats/formats.html.twig
@@ -22,7 +22,6 @@
   </div>
 </div>
 <script>
-
 function makeIdMap(json) {
   const out = new Map();
   json.data.forEach(d => {
@@ -34,7 +33,6 @@ function makeIdMap(json) {
 async function buildFormatsView() {
   const desiredFormats = new Set(['eternal', 'standard', 'startup']);
 
-  //const [snapshotsResponse, restrictionsResponse] = await Promise.all(
   const [r1, r2, r3, r4, r5] = await Promise.all(
     [
       fetch('{{ v3_api_url }}/api/v3/public/snapshots?filter[active]=true').then(data => data.json()),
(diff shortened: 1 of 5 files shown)


app/Resources/views/Formats/formats.html.twig
```diff
@@ -13,75 +13,118 @@
...
+      $(`#${format.id}_runner_decklists`).append(`<li><a href="${Routing.generate('decklists_list', {type:'find', 'faction':'criminal', 'packs': packs})}"><span class="icon icon-criminal influence-criminal"></span> Criminal</a></li>`);
+      $(`#${format.id}_runner_decklists`).append(`<li><a href="${Routing.generate('decklists_list', {type:'find', 'faction':'shaper', 'packs': packs})}"><span class="icon icon-shaper influence-shaper"></span> Shaper</a></li>`);
+      $(`#${format.id}_runner_decklists`).append(`<li><a href="${Routing.generate('decklists_list', {type:'find', 'faction':'adam', 'packs': packs})}"><span class="icon icon-adam influence-adam"></span> Adam</a></li>`);
+      $(`#${format.id}_runner_decklists`).append(`<li><a href="${Routing.generate('decklists_list', {type:'find', 'faction':'apex', 'packs': packs})}"><span class="icon icon-apex influence-apex"></span> Apex</a></li>`);
+      $(`#${format.id}_runner_decklists`).append(`<li><a href="${Routing.generate('decklists_list', {type:'find', 'faction':'sunny-lebeau', 'packs': packs})}"><span class="icon icon-sunny-lebeau influence-sunny-lebeau"></span> Sunny Lebeau</a></li>`);
```
distributive: Would it be possible to detect (or hardcode it) to exclude the mini-factions from formats that don't include them?

app/Resources/views/Formats/formats.html.twig
```diff
@@ -13,75 +13,118 @@
...
+
+      $(`#${format.id}`).append(`<ul id="${format.id}_cycles"></ul>`);
+      cycles.forEach(cycle => {
+        if (legalCycleIds.has(cycle.id)) {
+          $(`#${format.id}_cycles`).append(`<li><a href="${Routing.generate('cards_cycle', {cycle_code: cycle.attributes.legacy_code})}">${cycle.attributes.name}</a></li>`);
+        } else if (incompleteCycleIds.has(cycle.id)) {
+          const incompleteListId = `${format.id}_cycles_${cycle.id}`;
+          if (!$(`#${incompleteListId}`).length) {
+            $(`#${format.id}_cycles`).append(`<li>${cycle.attributes.name}<ul id=${incompleteListId}></ul></li>`);
+          }
```
distributive: Currently this only affects Terminal Directive in eternal, which makes it stand out kinda weird. Having all cycles show their sets would be way too long though so I don't have a good solution of the top of my head.

distributive (commented):
Would it be possible to add a few more comments between sections just to make it a little easier to see which section does what?

app/Resources/views/Formats/formats.html.twig
plural: i think i can tidy the JS code up a bit for this, but i like the presentation - it's similar to what we do in the collection settings for decklist search and the deckbuilder.

app/Resources/views/Formats/formats.html.twig
plural: sure.  easy enough since they are in one big box!

plural:
I'll make a pass over the code to add some more comments and see if there is some more stuff to deduplicate.  It's annoying to have all the HTML be buried in the JS, but such is life.

The config stuff is in now as well, so I think this is generally good (and then will be useful for further PRs as well).

plural:
allrighty, this should be good for a real review.

distributive (approved):

PR #688, opened 2022-09-11
Added password-reset console command

distributive:
Now we can reset passwords manually


plural:
ugh. i'm sorry.  when i patched this in and went to test it locally i saw that we *do* have a command for this, but it's provided by the framework, not in our code:
fos:user:change-password 

I should have checked the console instead of just our command folder.  :/

PR #689, opened 2022-09-12
New api ban list page

distributive:
- Updated the ban list page to dynamically load from the new API (requires the api to be running locally)
- Improved the layout of the ban list page (see below)
- Started a js file for general purpose v2api functions
- The ban list search query (b:...) now correctly interprets the inverse operator (b!...)

Some notes:
- The ban list search query seems to fail with older restriction types (universal faction cost, global penalty)
- Due to limitations on the api I haven't been able to work out, this version loads the entire card pool and then manually filters it for cards on each list (probably should not be made live until fixed)

image

image

image

image


plural:
sync in main and use the new config variable for the API.  We can get the API deployed shortly with your change for latest printing ID as well, but i am also going to take a crack at a couple of materialized views for cards and printings that may simplify things a bit as well.

distributive:
This still loads all card data every time the page is loaded, but that can be fixed when the api allows for it.

web/css/style.css
```diff
@@ -712,6 +712,8 @@ ul.rulings-list blockquote {
   list-style-type: '-  ';
 }
 
+
```
plural, line 753: do you need a whitespace-only change to this file?

web/js/api.js
```diff
@@ -0,0 +1,59 @@
+// Fetches all cards from the v2 API
+async function fetchCards() {
```
plural: something to consider until we have some proper ember stuff in place is what goes in an external JS file vs. a template included inline.

The way i did the formats page kept things inline at the end of the body without extra loads (except for the API calls obviously).

Since we don't have great (ha! any) cache busting for our JS, it might be good to not add too many new JS files to NRDB classic.

I would personally move this to a template we could include from the twig files side, I'm curious what @locks has to say here, though.

plural: also, YAY for starting to sort out some common client side libs we need.

web/js/api.js
```diff
@@ -0,0 +1,59 @@
...
+    data.push(...json.data);
+  }
+  return data;
+}
+
+// Remove all cards with any of the given subtypes from a list of cards
+function removeCardsWithSubtypes(cards, subtypes) {
+  if (!subtypes || subtypes.length == 0)
+    return cards;
+  return cards.filter(c => subtypes.some(s => !(c.attributes.display_subtypes && c.attributes.display_subtypes.toLowerCase().split(' - ').includes(s))));
```
plural: formatting nit, are there some more convenient line breaks / a lurking helper function here?

since we are only using this for currents right now, should this just be a "remove currents?" function?  I'd rather expose the subtype ids for operations rather than rely on the display subtypes.

plural: Also, you can include subtypes in the cards API call if you use https://api-preview.netrunnerdb.com/api/v3/public/cards/?include=card_subtypes

For Abagnale this adds:
```
"data": [
{
"type": "card_subtypes",
"id": "decoder"
},
{
"type": "card_subtypes",
"id": "icebreaker"
}
]
```

to relationships.card_subtypes

web/js/api.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// Searches a list of restrictions for the one with the given ID
+function getRestrictionById(restrictions, id) {
```
plural: consider what i did for the formats page?  https://github.com/NetrunnerDB/netrunnerdb/blob/main/app/Resources/views/Formats/formats.html.twig#L25

this method is not needed if the data is transformed into a map.

web/js/api.js
```diff
@@ -0,0 +1,59 @@
...
+    if (card.attributes.side_id == 'corp')
+      corp.push(card);
+    else
+      runner.push(card);
+  });
+  return [corp, runner];
+}
+
+// Generates a link to a card by creating a link to its most recent printing
+function cardToLink(card) {
```
plural: cardToLatestPrintingLink ?

src/AppBundle/Service/CardsData.php
```diff
@@ -719,8 +719,8 @@ public function get_search_rows(array $conditions, string $sortorder, string $lo
                         $mwl = $this->entityManager->getRepository(Mwl::class)->findOneBy(['code' => $condition[0]], ["dateStart" => "DESC"]);
                     }
                     if ($mwl) {
-                        // Exclude any cards banned by this ban list.
-                        $clauses[] = "(c.id NOT IN (SELECT mc.card_id FROM AppBundle:MwlCard mc WHERE mc.mwl_id = ?$i))";
+                        $cond = $operator == ":" ? "in" : "not in";
```
plural, line 722: this could be its own PR

app/Resources/views/Banlists/banlists.html.twig
```diff
@@ -2,88 +2,171 @@
...
-              </li> 
-            {% endfor %}
-          </ul>
-        </div>
-      </div>
-    {% endfor %}
-  </div>
 </div>
+
+<script>
```
plural: a couple of general things in here. 

would you please make the variables const where you can?
would you please make a pass through for formatting and indentation? There are some blocks that are pretty bunched up.

Also, could we do jquery dom manipulation here instead of string literals?  it would be good to be consistent for the pages we are changing.

plural (changes requested):
sorry for the pile of comments.

web/css/style.css
distributive: Oops! This is left over from a temporary addition to the file.

src/AppBundle/Service/CardsData.php
distributive: I can move it if you want, but technically this pr relies on it as otherwise there's no valid search to link to for cards affected by a banlist (although it still doesn't work for points or universal costs, etc)

web/js/api.js
distributive: Does the include let me get a explicit list of subtypes without making additional queries? I can't find one.

I think having the subtypes be exposed as a list that users have to join with " - " would make more sense to me than the `display_subtypes` we have now.

src/AppBundle/Service/CardsData.php
plural: ah, i didn't realize it was needed here.  Carry on.  :D

web/js/api.js
plural: yeah, take a look at the API link - include will include the ids for the related attributes and for things like card subtypes the id is the only thing you need.  this all happens in one http request to the API.

distributive: Ohhh I see! That's really useful, thanks!

app/Resources/views/Banlists/banlists.html.twig
```diff
@@ -1,89 +1,185 @@
...
+
+async function buildBanlistsView() {
+  // Add a temporary loading indicator
+  $('.tab-pane').append(
+    '<p class="temp-loading">Loading...</p>'
+  );
+
+  // Load data from API
+  const desiredFormats = ['startup', 'standard', 'eternal'];
+  const formats = await Promise.all(desiredFormats.map(f => fetch(`{{ v3_api_url }}/api/v3/public/formats/${f}`).then(data => data.json()).then(json => json.data)));
```
plural: couple of options for you here.

you don't need to grab the formats from the API.  

You could just call /api/v3/public/formats/${format}/restrictions OR just get all the restricitons in 1 API call and filter them here.

Additionally, you can sort in the API call: /api/v3/public/formats/standard/restrictions?sort=-date_start

app/Resources/views/Banlists/banlists.html.twig
```diff
@@ -1,89 +1,185 @@
...
+    const panels = $(this).closest('.list');
+    panels.find('.panel').children(':not(:first-child)').hide();
+    panels.find('.list-toggle').html('Show');
+  });
+
+  // Add the restriction data to each text box
+  for (const format of restrictions) {
+    for (const r of format) {
+      // Get the affected cards by restriction type
+      let cards = await fetchCards(`?include=card_subtypes&filter[search]=in_restriction:true restriction_id:${r.id}`).then(cards => splitBySide(cards));
```
plural: Fetching cards I think you can do in a single call.

/me checks some stuff

We'll need one other change to allow this to be done in a single call. I'll send you that PR in a minute.

The call you can make is http://localhost:3000/api/v3/public/cards?filter%5Bsearch%5D=in_restriction%3Atrue&page[limit]=250&page[offset]=0

we have a max page size of 1000 and a default of 100, but you can override it.

Since you have the restrictions and they have all the card ids, you could client-side slice these up from the single cards call.  I think this brings the number of requests down from 28 -> 4.

app/Resources/views/layout.html.twig
```diff
@@ -87,6 +88,7 @@
       gtag('config', 'UA-131671930-1');
     </script>
     {% endif %}
+    <script src="{{ asset('/js/api.js') }}"></script>
```
plural: you don't need this any longer.

plural (changes requested):

plural:
the layout looks great!  I would keep all but the most recent (or the active one) collapsed by default.

app/Resources/views/layout.html.twig
distributive: Oops, yeah!

app/Resources/views/Banlists/banlists.html.twig
distributive: Technically I need the formats for showing their names and setting some IDs, but given I've hardcoded in the three formats we want I'm happy to just hardcode those too

distributive: Is it better to make 1 call over 3? I'm happy to do either, I'm just wondering if there's a preference

app/Resources/views/Banlists/banlists.html.twig
distributive: No restriction has actually reached 100 cards yet (although it might one day if they ban another subtype)

I'll get it to manually slice these up though

distributive:
> the layout looks great! I would keep all but the most recent (or the active one) collapsed by default.

Sure thing!

If we're happy with this layout, in a later PR I'll separate the code into another library and start using it for other pages

app/Resources/views/Banlists/banlists.html.twig
distributive: Wait, I need the formats themselves to determine the active restriction 😐

distributive: If we're enforcing that restrictions only belong to one format then we can add an `active` flag to them themselves to get around this

plural: AH, right.  Names come from format calls.   A mirrored active flag for restrictions is a good idea - we can handle that separately, though.

I think your minimal set of requests is:
- get all formats
- get all restrictions
- get all cards on any restrictions

and all of those can happen in parallel.

we definitely should aim for the fewest # of requests we can get away with - it will be more responsive to the user and cause less load on the server.

app/Resources/views/Banlists/banlists.html.twig
```diff
@@ -35,6 +34,11 @@
       <div role="tabpanel" class="tab-pane" id="tab-pane-eternal"></div>
     </div>
   </div>
+
+  <noscript>
```
plural, line 38: the site is gonna be pretty sad without JS.  hehe

plural (approved):
This looks fantastic!

PR #694, opened 2022-09-27
Update formats page

distributive:
I hope you don't mind me playing around with your code @plural but I wanted to keep all the new pages up to date:
- Added the loading icon to the formats page (moving the html for it to a global location)
- Imported the new local api module and replaced `makeIdMap` with its counterpart
- Simplified the fetch commands with their counterparts


plural (approved):
looks great.

PR #695, opened 2022-09-28
New api illustrators page

distributive:
This adds a page listing every illustrator in the game and the printings they are responsible for. It also makes significant changes to the banlists page since the panels used there have been moved to their own module.


app/Resources/views/Illustrators/illustrators.html.twig
```diff
@@ -0,0 +1,60 @@
...
+  panelList.addSortOption("Name (A-Z)", (panel => panel.title), "ascending");
+  panelList.addSortOption("Name (Z-A)", (panel => panel.title), "descending");
+  panelList.addSortOption("Cards (most)", (panel => parseInt(panel.title.match(/\d+(?= card)/)[0])), "descending");
+  panelList.addSortOption("Cards (least)", (panel => parseInt(panel.title.match(/\d+(?= card)/)[0])), "ascending");
+  illustrators.forEach(illustrator => {
+    const panel = panelList.createPanel(illustrator.id, false);
+    const count = illustrator.relationships.printings.data.length;
+
+    const search = Routing.generate('cards_find', {type:'find', 'view':'list', 'q':`i:"${illustrator.attributes.name}"`});
+    panel.addHeader(`${illustrator.attributes.name} • (<a href="${search}">${count} ${count > 1 ? 'cards' : 'card'}</a>)`);
```
plural: the dot feels quite bold compared the illustrator names.  Maybe an em-dash instead?

app/Resources/views/Illustrators/illustrators.html.twig
```diff
@@ -0,0 +1,60 @@
...
+  const panelList = new PanelList($('#illustrators'), null, true, "search", "sort");
+  panelList.addSortOption("Name (A-Z)", (panel => panel.title), "ascending");
+  panelList.addSortOption("Name (Z-A)", (panel => panel.title), "descending");
+  panelList.addSortOption("Cards (most)", (panel => parseInt(panel.title.match(/\d+(?= card)/)[0])), "descending");
+  panelList.addSortOption("Cards (least)", (panel => parseInt(panel.title.match(/\d+(?= card)/)[0])), "ascending");
+  illustrators.forEach(illustrator => {
+    const panel = panelList.createPanel(illustrator.id, false);
+    const count = illustrator.relationships.printings.data.length;
+
+    const search = Routing.generate('cards_find', {type:'find', 'view':'list', 'q':`i:"${illustrator.attributes.name}"`});
```
plural: how about 'view':'images' ?

app/Resources/views/Illustrators/illustrators.html.twig
```diff
@@ -0,0 +1,60 @@
...
+    panel.addHeader(`${illustrator.attributes.name} • (<a href="${search}">${count} ${count > 1 ? 'cards' : 'card'}</a>)`);
+
+    // Format the illustrator's printings into 3 columns
+    panel.addBody();
+    panel.onShow = async function() {
+      const printings = await fetchData(`{{ v3_api_url }}/api/v3/public/illustrators/${illustrator.id}/printings`);
+      const list = printings.map(p => `<li>${printingToAnchor(p)}</li>`);
```
plural: the layout here is interesting.  I love the top level illustrator and sorting view.

When expanded, I expected to see the card images themselves. not just the names.

For most illustrators this is fine.  For Matt Z and a few others, it will be cumbersome.  While we could lazyily load the images when an illustrator is expanded, it will still be a lot for some illustrators.

wdyt?

app/Resources/views/Scripts/panels.html.twig
```diff
@@ -0,0 +1,224 @@
...
+  // Adds a sort option that sorts panels by passing them to valueFunc
+  static sortCount = 0;
+  addSortOption(desc, valueFunc, order="descending") {
+    const id = `sort-${PanelList.sortCount++}`;
+    const ascending = order == "ascending";
+    const caller = this;
+    this.sortDropdown.append(`<li><a href="#" onClick="return false;" id="${id}">${desc}</a></li>`);
+    this.sortDropdown.find(`#${id}`).on('click', function(event) {
+      event.stopPropagation();
+      caller.sortDropdown.dropdown("toggle");
```
plural, line 163: the sorting is not working when i patch this in.  Nothing obvious jumps out to me about why at this second, tho.

plural (changes requested):
this is exciting!  we should talk about layout a bit and the sorting wasn't working for me.

app/Resources/views/Illustrators/illustrators.html.twig
distributive: Trying to work out how to format it was a job. Initially I had it in the subheader but that meant it was hidden when the boxes were collapsed, and I think users would like to see how many printings each illustrator did.

distributive: I'll try it with an em-dash

app/Resources/views/Illustrators/illustrators.html.twig
distributive: Oh yeah good catch!

app/Resources/views/Illustrators/illustrators.html.twig
distributive: I thought about this and yeah, Matt Z et al made it hard to justify loading all the images. You can still see a thumbnail of each card when you hover them, but an alternative if that's not enough could be to switch to a list view that can also be expanded to show the art.

app/Resources/views/Illustrators/illustrators.html.twig
distributive: image

app/Resources/views/Scripts/panels.html.twig
distributive: Are you running it in Chrome? I just tried it in Chrome and it's not working any more (it does work in Firefox)

app/Resources/views/Illustrators/illustrators.html.twig
distributive: I found a better solution:
image

plural: YES.  this is great.  the on-demand loading works great and is really zippy.

plural (approved):

PR #696, opened 2022-09-30
Link up banlists from formats page.

plural:
The format will be selected and the banlist name will be pre-populated in the search box, if present.

Screen Shot 2022-09-29 at 8 50 54 PM
Screen Shot 2022-09-29 at 8 50 43 PM

I also replaced the windows line endings with unix line endings in the panel and banlist pages.


distributive (commented):
lol Windows is terrible.

This is good, the only issue is the search bar doesn't update the list until the user manually edits it. If this is the intended effect ignore this and merge.

distributive:
also do you know if we could extend this functionality to retain search data when you refresh? I don't think many people are going to be mashing F5 on the page but it's a nice QoL feature

app/Resources/views/Banlists/banlists.html.twig
```diff
@@ -94,13 +94,7 @@ async function buildBanlistsView() {
       formatRestrictions.forEach(r => {
         const active = r.id == f.attributes.active_restriction_id;
         let visible = active || r.id == formatRestrictions[0].id;
-        if (query_format && query_restriction) {
-          if (f.id == query_format && r.attributes.name == query_restriction) {
-            visible = true;
-          } else {
-            visible = false;
-          }
-        }
```
distributive: I'd be tempted to leave this in, so that if users empty the search bar only the banlist they were looking at before remains open.

distributive (approved):

app/Resources/views/Banlists/banlists.html.twig
plural: i wasn't happy with the way that ended up looking, but we can revisit it in a followup.

plural:
> also do you know if we could extend this functionality to retain search data when you refresh? I don't think many people are going to be mashing F5 on the page but it's a nice QoL feature

we should update the URL while we type as a follow-up and/or add copy icons to the individual rows in the table.

PR #697, opened 2022-09-30
Updated the name of the organisation

distributive:
Every instance of NISEI has been replaced (except when referring to cards, and in links that have not been updated yet).

Also fixed the line endings in the banlist page.


plural (approved):
we'll need to be careful not to revert the nisei glyph when we make new icons in the font file, but this is good.

PR #702, opened 2022-10-18
Fix filtering influence for identities

### LINKED ISSUE #692: Influence Search is behaving incorrectly
New umbrella issue for influence search shenanigans.

Other issues will be de-duped against this one.

muyjohno:
Example implementation to fix influence filter bug. I don't have a working local environment so can't vouch for this working, but wanted to offer it as an illustration.

Applying logic based on the code of the type rather than the ID makes it resistant to inconsistent IDs between environments.

Fixes #692

### DIFF
--- a/src/AppBundle/Service/CardsData.php
+++ b/src/AppBundle/Service/CardsData.php
@@ -566,16 +566,16 @@ class CardsData
                     foreach ($condition as $arg) {
                         switch ($operator) {
                             case ':':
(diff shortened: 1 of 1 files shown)


distributive (approved):
This was a really nice catch; thanks for working it out!

Incidentally it seems a bit counter intuitive that the draft IDs and other infinite-influence IDs are caught by searches for 0 influence, but I think that's something to solve when we redo the search algorithm for nrdb2.

plural (approved):
I patched this in and it works a treat.

🤦 🤦 🤦 

THANK YOU

PR #704, opened 2022-10-21
Cleaned up the syntax page

distributive:
- The syntax page now uses the new panels stuff (the content is still loaded server side, just reformatted locally)
- Criteria (e.g. `x:...` for text) are ordered alphabetically based on the property they sort
- The panel search bar now ignores html tags

### DIFF
--- a/app/Resources/views/Default/syntax.html.twig
+++ b/app/Resources/views/Default/syntax.html.twig
@@ -3,135 +3,480 @@
 {% block title %}NetrunnerDB Search Syntax{% endblock %}
 
 {% block body %}
+{% include '/Scripts/api.html.twig' %}
+{% include '/Scripts/panels.html.twig' %}
 <div class="container">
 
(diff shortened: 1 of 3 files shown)


distributive:
Also a note for prosperity: I accidentally pushed a slightly earlier version of this direct to main instead of my own fork, so that's why there's some similar commits before this (now reverted)

NoahTheDuke:
Looks like the line endings changed. Maybe we should change the .gitattributes to enforce consistency.

distributive:
I think my editor was setting things to Windows line endings for a while so that would be appreciated

plural (approved):
Looks great! Merge away!

NoahTheDuke:
Yeah, once the line endings are fixed, this looks great.

distributive:
I double checked and it seems the file originally had the wrong line endings and my new settings fixed it.

Issue #623, opened 2022-04-24
Punitive Counterstrike text shows old and new trace formatting on top of each other


infinitenutshell:
https://netrunnerdb.com/en/card/31078
New trace style (expected): Trace[5]. If successful...
Old trace style (still seen on cards that haven't been reprinted or added to the Card Text Updates document): **Trace [5] –** If successful...
Punitive currently displays as: **Trace [5] –**. If successful...

Probably there's just an extraneous bit of markup in the card data. This error does not currently appear for any other card.

Edit: fixed en dashes

plural:
Ah, interesting.

This isn't a problem with the card data.  This looks to be a rare instance (only?) of a Trace effect just being a full sentence.

The controller currently replaces `<trace>Trace 5</trace>` like this:
` $cardinfo['text'] = preg_replace('/<trace>([^<]+) ([X\d]+)<\/trace>/', '<strong>\1 [\2]</strong>–', $cardinfo['text']);`

To fix this, we would actually need to update all other traces in the JSON to add the en dash as a literal and then remove that en dash from the controller.

infinitenutshell:
I think you're overcomplicating it a bit. All you need to do is remove the `<trace>` tag that applies the old formatting. All these cards display the new formatting just fine: https://netrunnerdb.com/find/?q=x%3A%22trace%5B%22&sort=name&view=text&_locale=en

plural:
Those don't actually use the trace syntax.  They just have a literal Trace[#] in the text, like Executive Function.
Compare:
`    "text": "[subroutine] Trace[4]. If successful, do 1 brain damage.",`

to CtM's: 
` "text": "The first time the Runner trashes an installed Corp card each turn, you may <trace>trace 4</trace> If successful, give the Runner 1 tag (cannot be avoided).",`

plural:
incidentally, this also makes them normal weight and not bold.

plural:
if we have a trace syntax, why use it sometimes, but not other times?  Is trace a bit of text or an ability with a keyword/formatting?  Right now it's 🤷   It would be better for data quality to have it be a real keyword so we can match on the ability. The text formatting is pretty inconsistent now with respect to what is a keyword and what is some literal business shoved in (hello html tags grumble grumble).

infinitenutshell:
I noticed today that this issue now affects all trace cards updated in CTU1.3. Those cards are:

Power Grid Overload
Data Hound
Midseason Replacements
[...]

PR #705, opened 2022-10-25
Update trace text format

mankinchi:
Handling the styling issue mentioned in https://github.com/NetrunnerDB/netrunnerdb/issues/623. This would also allows some cards data change to work properly afterward

### DIFF
--- a/src/AppBundle/Service/CardsData.php
+++ b/src/AppBundle/Service/CardsData.php
@@ -888,7 +888,7 @@ class CardsData
         // If the card has text
         if (strlen($cardinfo['text']) > 0) {
             // replacing <trace>
-            $cardinfo['text'] = preg_replace('/<trace>([^<]+) ([X\d]+)<\/trace>/', '<strong>\1 [\2]</strong>–', $cardinfo['text']);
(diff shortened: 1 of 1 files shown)


plural (changes requested):
Please update to keep the match against the trace in the json as well.

Please also run locally and make sure things look ok.

src/AppBundle/Service/CardsData.php
```diff
@@ -888,7 +888,7 @@ public function getCardInfo(Card $card, string $locale)
         // If the card has text
         if (strlen($cardinfo['text']) > 0) {
             // replacing <trace>
-            $cardinfo['text'] = preg_replace('/<trace>([^<]+) ([X\d]+)<\/trace>/', '<strong>\1 [\2]</strong>–', $cardinfo['text']);
+            $cardinfo['text'] = preg_replace('/<trace>.*([X\d]+)<\/trace>/', '<span>Trace [\1]</span>.', $cardinfo['text']);
```
plural, line 891: please make the `.*` not greedy with `.*?` or use the original capture group here.

plural (changes requested):
just one small thing.

mankinchi:
so sorry for the disappearance. Personal issue is annoying but resolved. I will look back into this and update it asap 😄

plural:
I ended up taking care of this in #725 .  Thank you for kicking this off!

PR #708, opened 2022-11-10
Remove Jinteki.net mention from "Copy legacy URL" action

locks:
jnet has been updated to support both the legacy and the new URLs,
and someone mentioned on stimhack that legacy URLs are still needed for ABR.
I thought might as well make it generic.

I'll look into updating ABR as well.

### DIFF
--- a/app/Resources/views/Decklist/decklist.html.twig
+++ b/app/Resources/views/Decklist/decklist.html.twig
@@ -55,7 +55,7 @@ NRDB.user.params.decklist_id = Decklist.id;
                    }, function() {
                        alert('Copy to clipboard failed. Sorry!');
                    });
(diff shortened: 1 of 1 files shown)


distributive:
Do you know how many users actually know about the change in URLs? I'd maybe change it to _"Copy Legacy URL for Jinteki.net and ABR"_ so people know what the option is for.

NoahTheDuke:
We at jnet removed support for the legacy url a while ago and folks complained so we added it back, lol. I would love to not have to support it. I can do an update of jnet basically as soon as it's gone from nrdb.

distributive:
Oh in that case I don't mind especially

distributive (approved):

PR #710, opened 2022-11-20
Highlight first card suggestion during deckbuilding

vrumanuk:
Small QOL for deckbuilding which would allow for card selection without use of mouse/Tab/Down.

I manually tested this on a local instance and didn't see any issues but I'm unfamiliar with this stack and repo so let me know if there are any tests I should be running.


plural:
This looks good, thanks for the contribution.

For posterity, this changes the current behavior of:
Screen Shot 2022-11-20 at 11 29 06 PM

to
Screen Shot 2022-11-20 at 11 28 54 PM

Issue #700, opened 2022-10-03
Pre-Load/Cache Illustrators Page


GordonsBeard:
Currently the "Illustrators" page takes a few seconds to actually load and compile. I cannot imagine this page updates daily/weekly at all so I propose it would be better to create some kind of offline process to generate this page instead of trying to calculate it all on page load, online.

I could be wrong and this isn't really a "bug" more of a suggestion to reduce some load the website might experience given this page's length to load.

plural:
yeah it's looking slow.  :). We're doing some more work on the new API to improve performance so we'll keep an eye on this.

plural:
the fixes for this have been deployed.  Sorry for downloading ... checks notes ... 9.5MB before.  Down to 125KB.

Thanks for letting us know!

PR #711, opened 2022-11-25
Use new num_printings attribute from illustrators api.

plural:
This brings the illustrator load page (locally!) down from > 10s to < 100ms


plural:
This fixes #700

distributive (approved):

distributive:
Wait, don't merge yet

PR #715, opened 2022-12-09
Parhelion neutral ids

plural:


src/AppBundle/Service/Judge.php
```diff
@@ -248,6 +250,10 @@ public function analyse(array $slots)
                 $problem = 'copies';
             }
 
+            if (($identity->getCode() == '34093' /* Nova Initiumia */ || $identity->getCode() == '34128' /* Ampere */) && $qty > 1) {
```
NoahTheDuke: Should these be based on the name and not the code so it's forward compatible with any reprints?

plural: Good callout!

Now that the new API is in pretty good shape, we're moving on to deckbuilding and other things next.  That will all be card id based, so we'll sweep all that up when we do that.

distributive (approved):

PR #724, opened 2022-12-16
Add rounded corner CSS for card images.

plural:


web/css/style.css
```diff
@@ -6,6 +6,10 @@ body {font-size:16px;font-family: 'LatoWeb', sans-serif;}
 /* classes .card-* used in card pages, card modal, card tip */
 .pack-future, .card-preview{color:red}
 
+.card-image-rounded-corners {
```
distributive: I'm not sure how to link to it in this review but I added line 148 of this file a while back to do the same thing as this PR but in fewer places. It's probably best to remove that line (see https://github.com/NetrunnerDB/netrunnerdb/pull/657).

plural: Ah! Thanks for catching that.

I switched to using this with a tweak.

plural:
I also found and fixed the class and the alt text for images on the illustrators page.

distributive (approved):

PR #726, opened 2023-02-04
prefer prefix match in autocomplete

roitk:
Andrej from the Metropole Grid made a [valid point](https://www.youtube.com/live/R5F65deCDkU?feature=share&t=6300) that certain cards like Ping, Lat, and Ob will never appear in the autocomplete because we do a substring search of the card title and sort by title, and they get shoved way down in the list when you look for their full name. This changes the sort comparator to prefer cards that match on the prefix over those that just match elsewhere in the substring.

I'm having trouble testing this locally as the typeahead library seems to not be firing in my local Symfony web server (even on `main`). Is there something I have to set to get this to work in the local webserver?

### DIFF
--- a/app/Resources/views/Search/advanced-search.html.twig
+++ b/app/Resources/views/Search/advanced-search.html.twig
@@ -32,7 +32,7 @@
                 function normalizeTitle(cardTitle) {
                     return _.deburr(cardTitle).toLowerCase().trim();
                 }
-                var matchingCards = _.filter(all_cards, function (card) {
+                var matchingCards = _.filter(latest_cards, function (card) {
                     return regexp.test(normalizeTitle(card.title));
                 });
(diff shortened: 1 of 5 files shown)


roitk:
I've tried a couple of different local testing environments. Neither the local Symfony server nor the docker-compose setup from [nrdb-dev-env](https://github.com/Absotively/nrdb-dev-env) allow me to see the autocomplete activating in the browser. :(

EDIT: It was a permissions issue in my local setup, resolved now

roitk:
Working prefix match:
image

@plural brought up the point that some of the search bars unnecessarily show cards in multiple printings like so:

image

The logic to hide all but the latest printing actually already exists for a few of the search bars, but I'd rather save the fix for that for another PR to make this easy to review.

distributive (approved):
Thanks for implementing this!

PR #727, opened 2023-02-05
Update github URLs to point to netrunnerdb org.

plural:
The old Alsciende URLs are misleading.


distributive (approved):
That reminds me, we also need to update [this page](https://netrunnerdb.com/api/2.0/doc). I can do it tomorrow unless you happen to have time today.

PR #729, opened 2023-02-05
Updated API page to mention NSG + general cleanup

distributive:
Most notably the page now informs users of the existence of v3.


plural (approved):
Thank you!

PR #731, opened 2023-02-11
Added NSG-only as a deckbuilding format

distributive:
It includes every set that's every been legal in Startup.

image

System Core 19 and Salvaged Memories have been excluded as they were never released as a physical product. The Magnum Opus Reprint has been excluded because it is a direct replica of an existing FFG set.

### DIFF
--- a/web/js/deck.v2.js
+++ b/web/js/deck.v2.js
@@ -226,6 +226,27 @@ function create_collection_tab(initialPackSelection) {
     });
     update_collection_packs();
   });
+  $('#collection_nsg').on('click', function(event) {
+    let startup_cycles = Array(); // Hardcoded Startup Codes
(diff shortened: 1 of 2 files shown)


web/js/deck.v2.js
```diff
@@ -226,6 +226,27 @@ function create_collection_tab(initialPackSelection) {
     });
     update_collection_packs();
   });
+  $('#collection_nsg').on('click', function(event) {
+    let startup_cycles = Array(); // Hardcoded Startup Codes
+    startup_cycles['system-gateway'] = 1;
```
plural, line 231: when we have a publisher or designer field in the new API we could tidy this up.

czully:
Can you please add "NSG only" format to the Decklist search too? :)

PR #734, opened 2023-02-17
Made casual/Startup legality an explicitly format in the deckbuilder

distributive:
I actually did this a while back, but for some reason never PR'd it.

This is what the relevant section of the deckbuilder now looks like:

image

It hides the option for banlists if set to causal:

image

### DIFF
--- a/app/Resources/views/Builder/deck.html.twig
+++ b/app/Resources/views/Builder/deck.html.twig
@@ -100,13 +100,16 @@ var Filters = {},
   </div>
   <div class="form-group">
     <label for="mwl_code">Legality</label>
-    <select class="form-control" name="mwl_code" id="mwl_code">
+    <select class="form-control" name="format_code" id="format_code">
+      <option value="standard" selected="selected">Standard</option>
+      <option value="casual">Startup/Casual</option>
+    </select>
+    <select class="form-control" name="mwl_code" id="mwl_code" style="margin-top: 10px;">
       {% for mwl in list_mwl %}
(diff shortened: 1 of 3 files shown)


app/Resources/views/Builder/deck.html.twig
```diff
@@ -100,13 +100,16 @@ var Filters = {},
   </div>
   <div class="form-group">
     <label for="mwl_code">Legality</label>
-    <select class="form-control" name="mwl_code" id="mwl_code">
+    <select class="form-control" name="format_code" id="format_code">
```
plural: it may be worth a note somewhere that says this is just for the builder and not saved. Could just be a reminder in the release notes/updates file.

plural (approved):

app/Resources/views/Builder/deck.html.twig
distributive: image
image

distributive: Wait, that's not true actually; decklists _do_ save MWL data. I've updated the backend to check if the format has been set to casual and record the MWL as `null` in that case.

I've also changed the layout a bit since it felt weird do have a dropdown with only two options:

image

distributive: Hopefully renaming the section to "Banlist" will make it clear that this isn't an implementation of format, but I'll also add a note to the update log when it goes live.

plural (approved):

PR #748, opened 2023-03-25
remove escaping from card flavor text

### LINKED ISSUE #753: HTML tags are visible in CBI Raid flavor text
See https://netrunnerdb.com/en/card/10022

The same applies to all cards containing HTML tags in their flavor text.

francescopellegrini:
Fixes NetrunnerDB/netrunner-cards-json#753


distributive (changes requested):
The reason this was broken is because there is now flavour text with html tags in it:
image

This fix reverts every other card, but entirely breaks this one page.

PR #759, opened 2023-07-22
stop user submitted img to overflow container

mertenhanisch:
When users submit images via md they can be bigger than their container (ie. current DOTW, 07/22/2023). This PR sets the max-width of images in the .decklist container without changing anything else about them.


plural:
Nice, thank you!

PR #778, opened 2023-11-21
Quick fix for spambots

distributive:
Requires users to have at least one private decklist before posting reviews/comments

image
image

### DIFF
--- a/src/AppBundle/Controller/ReviewController.php
+++ b/src/AppBundle/Controller/ReviewController.php
@@ -258,6 +258,7 @@ class ReviewController extends Controller
         }
 
         $user = $this->getUser();
+        $userVerified = $user ? $user->isVerified() : false;
 
         return $this->render(
 
@@ -275,7 +276,7 @@ class ReviewController extends Controller
                 'nexturl'         => $currpage == $nbpages ? null : $this->generateUrl($route, $params + [
                         "page" => $nextpage,
                     ]),
-                'comments_enabled' => $user->isVerified(),
+                'comments_enabled' => $userVerified,
             ],
 
             $response
--- a/src/AppBundle/Controller/ReviewController.php
+++ b/src/AppBundle/Controller/ReviewController.php
@@ -38,7 +38,7 @@ class ReviewController extends Controller
         }
 
         // bot prevention - this shouldn't ever happen unless a user messes with the comment input source
-        if (count($user->getDecks()) > 0) {
+        if (!$user->isVerified()) {
             return new Response(json_encode("You must have at least one private decklist before you may write a review."));
         }
 
@@ -275,7 +275,7 @@ class ReviewController extends Controller
                 'nexturl'         => $currpage == $nbpages ? null : $this->generateUrl($route, $params + [
                         "page" => $nextpage,
                     ]),
-                'comments_enabled' => count($user->getDecks()) > 0,
+                'comments_enabled' => $user->isVerified(),
             ],
 
             $response
@@ -359,7 +359,7 @@ class ReviewController extends Controller
                         "user_id" => $user->getId(),
                         "page"    => $nextpage,
                     ]),
-                'comments_enabled' => count($user->getDecks()) > 0,
+                'comments_enabled' => $user->isVerified(),
             ],
 
             $response
@@ -380,7 +380,7 @@ class ReviewController extends Controller
         $user = $this->getUser();
 
         // bot prevention - this shouldn't ever happen unless a user messes with the comment input source
-        if (count($user->getDecks()) > 0) {
+        if (!$user->isVerified()) {
             return new Response(json_encode("You must have at least one private decklist before you may post comments."));
         }
 
--- a/src/AppBundle/Controller/ReviewController.php
+++ b/src/AppBundle/Controller/ReviewController.php
@@ -37,6 +37,11 @@ class ReviewController extends Controller
             return new Response(json_encode("Your reputation doesn't allow you to write more reviews."));
         }
 
+        // bot prevention - this shouldn't ever happen unless a user messes with the comment input source
+        if (count($user->getDecks()) > 0) {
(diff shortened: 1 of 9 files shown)


src/AppBundle/Controller/ReviewController.php
```diff
@@ -37,6 +37,11 @@ public function postAction(Request $request, EntityManagerInterface $entityManag
             return new Response(json_encode("Your reputation doesn't allow you to write more reviews."));
         }
 
+        // bot prevention - this shouldn't ever happen unless a user messes with the comment input source
+        if (count($user->getDecks()) > 0) {
```
NoahTheDuke: if number of decks is greater than 0 (meaning they have decks), exit early?

distributive: Whoops!

src/AppBundle/Controller/SearchController.php
```diff
@@ -633,6 +635,7 @@ public function displayAction(
             "metadescription" => $meta,
             "locales"         => $locales,
             "currentRotationCycles" => $currentRotationCycles,
+            "comments_enabled"      => $user->isVerified(),
```
plural: this needs to handle users who are not logged in.

$user ? $user->isVerified() : false

src/AppBundle/Entity/User.php
```diff
@@ -328,6 +328,14 @@ public function isSupporter()
         return ($this->donation > 0 || $this->patreon_pledge_cents > 0);
     }
 
+    /**
+     * @return boolean True iff the user has at least one private decklist, proving they're not a bot
```
plural, line 332: s/proving/suggesting hopefully please oh please/. ;)

src/AppBundle/Service/PersonalizationHelper.php
```diff
@@ -51,6 +51,7 @@ public function defaultBlock(User $user)
             'unchecked_activity' => $this->activityHelper->countUncheckedItems($this->activityHelper->getItems($user)),
             'is_moderator'       => $this->authorizationChecker->isGranted('ROLE_MODERATOR'),
             'roles'              => $user->getRoles(),
+            'comments_enabled'   => $user->isVerified(),
```
plural, line 54: $user ? $user->isVerified() : false

src/AppBundle/Controller/ReviewController.php
```diff
@@ -268,6 +275,7 @@ public function listAction(int $page = 1, Request $request, EntityManagerInterfa
                 'nexturl'         => $currpage == $nbpages ? null : $this->generateUrl($route, $params + [
                         "page" => $nextpage,
                     ]),
+                'comments_enabled' => $user->isVerified(),
```
plural: $user ? $user->isVerified() : false

src/AppBundle/Controller/ReviewController.php
```diff
@@ -351,6 +359,7 @@ public function byauthorAction(User $user, int $page = 1, Request $request, Enti
                         "user_id" => $user->getId(),
                         "page"    => $nextpage,
                     ]),
+                'comments_enabled' => $user->isVerified(),
```
plural: $user ? $user->isVerified() : false

src/AppBundle/Controller/ReviewController.php
```diff
@@ -370,6 +379,11 @@ public function commentAction(Request $request, EntityManagerInterface $entityMa
     {
         $user = $this->getUser();
 
+        // bot prevention - this shouldn't ever happen unless a user messes with the comment input source
+        if (!$user->isVerified()) {
+            return new Response(json_encode("You must have at least one private decklist before you may post comments."));
```
plural, line 389: honestly, i might just pretend it worked instead of returning this error.

plural (commented):
There are definitely a few places where you need to be defensive but i might have been over-cautious with one or two.

src/AppBundle/Service/PersonalizationHelper.php
NoahTheDuke: Is this one needed? You're using user directly above.

src/AppBundle/Controller/ReviewController.php
NoahTheDuke: Same

src/AppBundle/Controller/SearchController.php
distributive: This flag doesn't actually enable comments any time they wouldn't already be, but I'll make the change for clarity

src/AppBundle/Controller/ReviewController.php
distributive: This _should_ only show up if the user's playing around with the inspector, but on the offchance this does happen naturally this will help users help us identify the issue.

src/AppBundle/Service/PersonalizationHelper.php
distributive: Yeah, in these cases multiple fields assume `$user` exists and gets it from an argument rather than fetching the user itself, so I'm happy to just assume it's fine

src/AppBundle/Controller/ReviewController.php
plural: this one 100% crashes .  you get the user above, but it can be null.  This one isn't annotated to enforce that a user is authenticated:

Screenshot 2023-11-22 at 8 59 47 AM

src/AppBundle/Controller/ReviewController.php
plural: This one is the user id of the author of the comments, not the user trying to make comments.  While this one shouldn't be null, this is checking against the wrong user.

src/AppBundle/Controller/SearchController.php
plural: sure, but this one will crash as well.  $user may be null.

plural (approved):
thanks for pulling this together! Let's see how it does.

PR #779, opened 2023-11-21
Fixed tags being added to review comments

### LINKED ISSUE #743: Reviews comments formatting
In the reviews page the comments appear with a <p> tag

https://netrunnerdb.com/en/reviews

imagen

distributive:
This doesn't fix the issue of comments being saved with their special chars sanitised (meaning Netrunner symbols render as `<span class="icon icon-subroutine"></span>` in comments).

Partially fixes https://github.com/NetrunnerDB/netrunnerdb/issues/743

### DIFF
--- a/app/Resources/views/Reviews/reviews.html.twig
+++ b/app/Resources/views/Reviews/reviews.html.twig
@@ -60,7 +60,7 @@
        {% if review.comments|length %}
 
        {% for comment in review.comments %}
-       <div class="review-comment text-muted small">{{ comment.text }}
+       <div class="review-comment text-muted small">{{ comment.text|raw }}
        &mdash;
        <a title="{{ comment.author.reputation }} reputation" href="{{ path('user_profile_view', {user_id:comment.author.id,user_name:comment.author.username|e('url')}) }}" rel="author" class="username {{ comment.author.faction }}">{{ comment.author.username }}</a>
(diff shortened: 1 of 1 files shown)


app/Resources/views/Reviews/reviews.html.twig
```diff
@@ -60,7 +60,7 @@
        {% if review.comments|length %}
 
        {% for comment in review.comments %}
-       <div class="review-comment text-muted small">{{ comment.text }}
+       <div class="review-comment text-muted small">{{ comment.text|raw }}
```
plural, line 63: i'm confused by this one.  This will not do any escaping.  Anything that was already saved in teh db that has unescaped tags will just be plopped out onto the page.  What am I missing?

distributive: With:
image

Without:
image

plural: sure, but if someone puts a script tag in there, it's gonna be written out to the page and execute.

The comment version already seems correct in that it doesn't create an xss path.  are the comments formatted as HTML when being saved to the database?

distributive: They get escaped, that's why icons appear as span tags
image

plural (approved):

PR #780, opened 2023-11-21
Flavour text HTML tags fix

distributive:
- Fixed html tags being escaped in flavour texts
- Spark of Inspiration has been hardcoded to not break the card page layout

### DIFF
--- a/app/Resources/views/Search/display-zoom.html.twig
+++ b/app/Resources/views/Search/display-zoom.html.twig
@@ -87,9 +87,9 @@
                 {% endif %}
 
                 {% if card.code == "33084" %}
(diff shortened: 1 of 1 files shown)


app/Resources/views/Search/display-zoom.html.twig
```diff
@@ -86,10 +86,14 @@
                     </div>
                 {% endif %}
 
-                {% if card.flavor %}
+                {% if card.code == "33084" %}
```
plural: if you wanted to move the conditional into the div content it would cut down on a little duplication.

plural (approved):

PR #788, opened 2023-12-29
Dark mode

### LINKED ISSUE #435: Feature Request: Dark Mode
There are Firefox Plugins like [Dark Reader](https://github.com/darkreader/darkreader) that enable you to dynamically create dark themes for any website. It does a pretty good job on NRDB (see screenshots attached), but has two main issues: Influence pips between Weyland and Neutral cards are indistinguishable. Also buttons are not easily readable in their on/off state (especially those showing how many copies of a card you include in your deck).
It would be cool to have a well-made dark mode option inbuilt to NRDB.

image
image

AlexRodwell:
Add basic dark mode support via [`@media (prefers-color-scheme: dark)`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).

Addresses feature request: https://github.com/NetrunnerDB/netrunnerdb/issues/435

### DIFF
--- a/web/js/nrdb.tip.js
+++ b/web/js/nrdb.tip.js
@@ -32,7 +32,7 @@
             type += '<p>Strength <b>' + card.strength + '</b></p>';
         var image_svg = '';
         if($('#nrdb_svg_hex').length) {
-            image_svg = '<div class="card-image-wrapper"><div class="card-image card-image-' + card.side_code + '-' + card.type_code + '"' + (card.imageUrl ? ' style="background-image:url(' + NRDB.card_image_url + '/small/' + card.code+ '.jpg)"' : '') + '></div></div>';
+            image_svg = '<div class="card-image-wrapper"><div class="card-image-mask"><div class="card-image card-image-' + card.side_code + '-' + card.type_code + '"' + (card.imageUrl ? ' style="background-image:url(' + NRDB.card_image_url + '/small/' + card.code+ '.jpg)"' : '') + '></div></div></div>';
         }
 
         $('.qtip').each(function(){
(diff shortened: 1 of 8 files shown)


web/js/nrdb.tip.js
AlexRodwell, line 1: Notably, I removed the <svg> mask used for the hexagon in the NRDB qtip here, it has been replaced with a CSS clip-path (widely supported) to better handle dark mode, as it previously would always have a white background.

distributive:
This looks really good!
I found a few more places with light mode left intact:
- Code blocks (see the search syntax page)
- Some buttons on the private decklists page
- Some buttons on the deckbuilder (including when you mouseover cards in the build tab)
- Modals (decklist publishing, card popups, etc)

image

image

image

image

image

AlexRodwell:
Thanks @distributive, I believe there's also some other modals I missed on top of the decklist publish one you've noted. So if you find anything else, please flag it and I'll get it sorted when I can.

AlexRodwell:
My last change should cover the generic buttons/modals, as well as other elements that were flagged by @distributive.

I opted to leave Bootstraps warning styling to its default, as making warnings/errors appear in dark mode with the same emphasis can be difficult (debatable subject).

distributive:
_One last thing:_ (spotted in the syntax page)

image

Otherwise it all looks good!

AlexRodwell:
> _One last thing:_ (spotted in the syntax page)
> 
> image
> 
> Otherwise it all looks good!

Fixed in my last commit

distributive (approved):

PR #802, opened 2024-01-09
Improve accessibility of fallback text values in card details

### LINKED ISSUE #787: Subroutine arrows (“↳”) in card page not copy-pasteable
What looks like `↳ End the run.` in html is `subroutine End the run.` in plaintext, and thus gets copied as the latter.

Suggestions (either):

1. Use `↳`: That is, replace subroutine arrow images with the Unicode character *Downwards Arrow with Tip Rightwards*.
2. Use `<abbr>` tags (for unicode plaintext) with `background-image:`’d classes (for subroutine arrow images).

The latter is better (it works for all symbols, like credits), but the former seems easier (and may work for a few symbols, which have Unicode lookalikes).

For example, in *Scryfall*, the two symbols in [Llanowar Elves](https://scryfall.com/card/lea/210/llanowar-elves)’s text can be highlighted and copied as `{T}: Add {G}.`, but is still rendered correctly. It does so with `<abbr class="card-symbol card-symbol-T" title="tap this permanent">{T}</abbr>` in source, where `.card-symbol-T { background-image: url("data:image/svg+xml;base64, … ") }` and `.card-symbol { display: inline-block; overflow: hidden; width: 15px; height: 15px; cursor: help; … }`.

AlexRodwell:
Updated `replaceSymbols()` function to use a singular `<abbr>` with a `title` attribute, instead of a secondary `<span>` fallback item (fallback valued moved inside `<abbr>`, and CSS added to handle sizing of icons).

Resolves https://github.com/NetrunnerDB/netrunnerdb/issues/787.

### DIFF
--- a/app/Resources/views/Search/display-zoom.html.twig
+++ b/app/Resources/views/Search/display-zoom.html.twig
@@ -32,7 +32,7 @@
                             {{ card.advancementcost }}/{{ card.agendapoints }}<span class="icon icon-agenda-points"></span>
                         {% endif %}
                         {% if card.type_code != "identity" and card.type_code != "agenda" %}
-                            <abbr title="{{ card.cost }} Credit{{ card.cost > 1 ? 's' : '' }}">{{ card.formatted_cost|raw }}</abbr>
+                            <abbr title="{{ card.cost }} Credit{{ card.cost != 1 ? 's' : '' }}">{{ card.formatted_cost|raw }}</abbr>
                         {% endif %}
                     </span>
                 </h3>
--- a/app/Resources/views/Search/display-zoom.html.twig
+++ b/app/Resources/views/Search/display-zoom.html.twig
@@ -32,7 +32,7 @@
                             {{ card.advancementcost }}/{{ card.agendapoints }}<span class="icon icon-agenda-points"></span>
                         {% endif %}
                         {% if card.type_code != "identity" and card.type_code != "agenda" %}
-                            {{ card.formatted_cost|raw }}
+                            <abbr title="{{ card.cost }} Credit{{ card.cost > 1 ? 's' : '' }}">{{ card.formatted_cost|raw }}</abbr>
                         {% endif %}
                     </span>
                 </h3>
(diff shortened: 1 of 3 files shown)


AlexRodwell:
firefox_kXsqr8DBpd
firefox_yWdniPte3I

Copy/pasted text from Starlight Knight:

> Starlit Knight 5[credit]
> Ice: Sentry - Observer
> Strength: 2
> Influence: 3
> 
> Threat 4 → When the Runner encounters this ice, it gains X "[subroutine] End the run." subroutines for the remainder of this run, after its other subroutines. X is equal to the number of tags the Runner has.
> 
> [subroutine] Give the Runner 1 tag.
> 
> [subroutine] Give the Runner 1 tag.
> An exemplar of Starlightʼs hubris, it stands vigil over secrets of hypocrisy.
> Illustrated by [Jakuza](http://localhost:8080/app_dev.php/find/?type=find&_locale=en&view=images&q=i%3A%22Jakuza%22)

app/Resources/views/Search/display-zoom.html.twig
```diff
@@ -32,7 +32,7 @@
                             {{ card.advancementcost }}/{{ card.agendapoints }}<span class="icon icon-agenda-points"></span>
                         {% endif %}
                         {% if card.type_code != "identity" and card.type_code != "agenda" %}
-                            {{ card.formatted_cost|raw }}
+                            <abbr title="{{ card.cost }} Credit{{ card.cost > 1 ? 's' : '' }}">{{ card.formatted_cost|raw }}</abbr>
```
NoahTheDuke: Maybe `card.cost != 1 ? 's' : ''` to handle "0 Credits" as well?

NoahTheDuke (approved):

app/Resources/views/Search/display-zoom.html.twig
AlexRodwell: Good catch, will update!

NoahTheDuke:
I'll let the Null Signal folks actually merge this (even tho I still have write access) cuz it's their domain now, but I think this looks great and I'm very glad someone with experience with screen readers is making it better.

distributive (approved):
This is great. I'd love to extend this somewhere far down the line to give user's info about keywords and subtypes in card text.

PR #810, opened 2024-03-02
Display pronouns on card pages, when applicable (using v3 api)

distributive:


plural (approved):

app/Resources/views/Search/display-zoom.html.twig
```diff
@@ -8,6 +8,19 @@
 {% endblock %}
 
 {% block main %}
+{% include '/Scripts/api.html.twig' %}
+<script type="text/javascript">
+    async function getPronouns() {
+        // Attempt to get card's pronouns
+        const printing = await fetchData(`${v3_api_url}/api/v3/public/printings/{{ card.code }}`);
```
plural, line 15: oh, you shouldn't need both of these fetches.  Printing has the pronouns already.

plural: also, you need to pluck the pronouns out of the attributes element (from card or printing as you settle on that).

PR #823, opened 2024-06-05
Pronunciation Front End 

talha-ahsan:
adds a little section below the illustrator information which contains pronunciation guide info when provided

currently the cards json have pronunciation info for borealis and liberation cycles.


app/Resources/views/Default/about.html.twig
```diff
@@ -20,5 +20,11 @@
...
   <p><b>Governance</b>: NetrunnerDB is not a right. The word of the Moderation &amp; Administration team is final. Please do not attempt to harass or intentionally annoy either group. However, if you feel you've been treated unfairly by a Moderator or Administrator, please let us know, and let's talk about it.</p>
+
+  <h2 id="pronunciation-disclaimer"> A Note on Pronunciations </h2>
+  <p>NetrunnerDB provides pronunciation information when available for cards designed by Null Signal Games. This information is available in both IPA form, and an approximation designed for English speakers. There may be some bias in the pronunciation featured on this website for a card, as the original guides are prepared by an Australian with their accent not necessarily matching your own.</p> 
+
+  <h3 id="borealis-pronunciation-note">A Note on Borealis Cycle Pronunciations</h3>
```
distributive, line 27: Is this not also true of Liberation?

plural (approved):

app/Resources/views/Default/about.html.twig
talha-ahsan: This language was originally in the main section, but I was asked by Narrative to make it a specific note for Borealis.

PR #846, opened 2024-09-29
Add html sanitisation to front and backend

distributive:
No guarantee this hits everything, but it should prevent any new decklists/reviews/review comments having xss.

### DIFF
--- a/web/js/sanitize.js
+++ b/web/js/sanitize.js
@@ -1,0 +1,13 @@
+function escapeHtml(text) {
+  let map = {
+    '&': '&amp;',
+    '<': '&lt;',
+    '>': '&gt;',
+    '"': '&quot;',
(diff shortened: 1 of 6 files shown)


web/js/sanitize.js
```diff
@@ -0,0 +1,13 @@
+function escapeHtml(text) {
+  let map = {
+    '&': '&amp;',
+    '<': '&lt;',
+    '>': '&gt;',
+    '"': '&quot;',
+    "'": '&#039;'
+  };
+
+  console.log(text.replace(/[&<>"']/g, function(m) { return map[m]; }));
```
NoahTheDuke: intended to log?

distributive: Nope. Oops!

plural (approved):

PR #847, opened 2024-09-29
Identity selection improvements

distributive:
It now looks like this:
image


distributive:
Accidentally branched from the sanitation branch. I'm not fixing that.

plural:
Thanks for this! I made one change to add Faction icons here as well.

Screenshot 2024-09-30 at 12 47 31 AM
Screenshot 2024-09-30 at 12 47 16 AM

PR #848, opened 2024-09-29
Made plaintext decklist outputs more friendly to screenreaders

distributive:
Previously the plaintext data for decklists would include an `x` before each card (e.g. "3x Sure Gamble") which screenreaders would read out, making it an inaccessible feature. It also reads out the influence pips individually, so those have been simplified to plaintext.

image

### DIFF
--- a/web/js/nrdb.js
+++ b/web/js/nrdb.js
@@ -549,7 +549,7 @@ function check_ampere_agenda_limits() {
 
 function check_startup_constraints() {
     if(MWL && MWL['name'].includes("Startup Ban List 24.09")){
-        const five_threes = NRDB.data.cards.find({ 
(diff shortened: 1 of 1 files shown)


distributive:
My editor added that extra space after the five_threes line. I don't know why.

plural:
This makes it better for screen readers and worse for browser users.  Let's take a look at how we can get the best of both worlds here.

distributive:
Feel free to edit this

migueldlr:
Hey @distributive ! We're re-reviewing this issue. Could you share a screen recording of the current inaccessible behavior? It'd help us better understand what this PR is doing

distributive:
Go to the plaintext output of a deck:
image

Note that there is an "x" after the number of each card, that a screen reader will read out:
image

PR #849, opened 2024-09-29
Removed update log

distributive:
That thing doesn't appear to have been updated in a year. Soooo, if you want it gone, this does that.


plural:
I think we should get in the habit of updating this and probably pull it into the repo instead of unmanaged on the server.

PR #850, opened 2024-09-29
Made the decklist option in the navbar a dropdown

distributive:
In retrospect, this was obvious.
image

I also reordered the decklist categories based on how useful they are (importantly search is now at the top)
image


plural:
this is lovely, thanks!

PR #851, opened 2024-10-03
Add back in faction icons to identity selector.

plural:
I thought i added this in #847 but i clearly don't know how git works.

Screenshot 2024-09-30 at 12 47 31 AM
Screenshot 2024-09-30 at 12 47 16 AM


distributive (approved):
If you add `style="transform: translateY(2px);"` to the faction icons they'll line up better with the text.

plural:
> If you add `style="transform: translateY(2px);"` to the faction icons they'll line up better with the text.

ooh, thanks!  That is much nicer.

PR #859, opened 2024-10-12
Re-order decklist navigation items.

plural:


plural:
Andrej had some valid feedback on stream this week, so i re-ordered these items.

PR #868, opened 2024-11-22
Sanitize instead of escape

### LINKED ISSUE #864: Text encoding errors in decklist notes/descriptions
Text encoding is broken after I edit a decklist note. For example, this one was fine until I edited it recently. I tested another decklist note, and the text was fine there until I edited it.

Screenshot 2024-11-01 140645

xdg:
PR #846 escaped special HTML characters in raw user-provided text, such as deck notes, with the unintentional side effect that these entities were encoded on save, and then re-encoded on the next save, etc. leaving deck notes, etc. effectively un-editable from a practical perspective.

This PR reverts #846 and replaces it with HTML sanitization using DOM Purify when raw user-provided text is rendered via Javascript.

Fixes #864.


src/AppBundle/Controller/BuilderController.php
```diff
@@ -674,7 +674,7 @@ public function saveAction(Request $request, EntityManagerInterface $entityManag
         }
         $name = filter_var($request->get('name'), FILTER_SANITIZE_STRING, FILTER_FLAG_NO_ENCODE_QUOTES);
         $decklist_id = intval(filter_var($request->get('decklist_id'), FILTER_SANITIZE_NUMBER_INT));
-        $description = filter_var(trim(($request->get('description'))), FILTER_SANITIZE_SPECIAL_CHARS);
+        $description = trim($request->get('description'));
```
plural: We can't trust the client to always send us good data.   The JS side stuff is helpful, but we'll need server side sanitization as a backstop.

plural: We have HTMLPurifier available on the backend via the TextProcessor service that we can use for this.

```diff
--- a/src/AppBundle/Controller/BuilderController.php
+++ b/src/AppBundle/Controller/BuilderController.php
@@ -12,6 +12,7 @@ use AppBundle\Entity\User;
 use AppBundle\Service\CardsData;
 use AppBundle\Service\DeckManager;
 use AppBundle\Service\Judge;
+use AppBundle\Service\TextProcessor;
 use Doctrine\ORM\EntityManagerInterface;
 use Psr\Log\LoggerInterface;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
@@ -630,11 +631,12 @@ class BuilderController extends Controller
      * @param Request                $request
      * @param EntityManagerInterface $entityManager
      * @param DeckManager            $deckManager
+     * @param TextProcessor          $textProcessor
      * @return \Symfony\Component\HttpFoundation\RedirectResponse|Response
      *
      * @IsGranted("IS_AUTHENTICATED_REMEMBERED")
      */
-    public function saveAction(Request $request, EntityManagerInterface $entityManager, DeckManager $deckManager)
+    public function saveAction(Request $request, EntityManagerInterface $entityManager, DeckManager $deckManager, TextProcessor $textProcessor)
     {
         /** @var User $user */
         $user = $this->getUser();
@@ -674,7 +676,7 @@ class BuilderController extends Controller
         }
         $name = filter_var($request->get('name'), FILTER_SANITIZE_STRING, FILTER_FLAG_NO_ENCODE_QUOTES);
         $decklist_id = intval(filter_var($request->get('decklist_id'), FILTER_SANITIZE_NUMBER_INT));
-        $description = trim($request->get('description'));
+        $description = $textProcessor->purify(trim($request->get('description')));
         $tags = explode(',', filter_var($request->get('tags'), FILTER_SANITIZE_STRING, FILTER_FLAG_NO_ENCODE_QUOTES));
         $mwl_code = $request->get('format_casual') ? null : $request->get('mwl_code');

```

should do it, but i think we need something similar on decklist publishing as well.  You could do that here or we can do a followup.

plural:
Thanks! One other thing is that the preview does not appear to be working with this PR.  While typing deck notes, the preview box stays static:
Screenshot 2024-11-23 at 9 05 17 AM

plural (changes requested):

xdg:
The preview box not updating until keypress is like that on the deployed nrdb already.

src/AppBundle/Controller/BuilderController.php
xdg: OK, I'll make that change.  Published decks get sanitized via `markdown` here: https://github.com/NetrunnerDB/netrunnerdb/blob/main/src/AppBundle/Controller/SocialController.php#L129

plural:
Update: Locally cached JS was the problem.  Notes editing preview is a-ok.

src/AppBundle/Controller/BuilderController.php
plural: Ah.  i had locally cached JS.  Clearing the cache makes the preview work.

plural (approved):

Phamduy1247:
app/Resources/views/Banlists/banlists.html.twig

plural:
> app/Resources/views/Banlists/banlists.html.twig

?

PR #871, opened 2025-01-11
Very basic support for cards with multiple faces.

plural:
https://github.com/user-attachments/assets/391599bf-4152-437d-b0e2-1d2fb1d1304e

https://github.com/user-attachments/assets/b82316ee-ce2d-4196-b1b8-722ed57d2684

https://github.com/user-attachments/assets/0c0f1439-7df7-4b7f-ae5e-c1e31a16470d

### DIFF
--- a/app/Resources/views/Search/display-zoom.html.twig
+++ b/app/Resources/views/Search/display-zoom.html.twig
@@ -21,55 +21,55 @@
             ]
         );
 
-        // Have a single button to flip if there are 2 sides.
-        if (printing.data.attributes.num_extra_faces == 1) {
+        if (printing.data.attributes.num_extra_faces > 0) {
+            // Make a list of face buttons for cards with more than 1 face.
             const container = document.getElementById('card_image_container');
-            const div = document.createElement('div');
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
That looks great.

TotsuzenheniYukimi:
For cards that are one sided and have multiple faces because there are multiple versions of that one sided card, i suggest making the 'Front' face 1, perhaps naming it 'Face 1' or 'Front Face 1' and naming the next face 'Face 2' instead of 'Face 1', and so on.

plural:
This actually has lovely little animations now thanks to @AlexRodwell !!!

plural:
> For cards that are one sided and have multiple faces because there are multiple versions of that one sided card, i suggest making the 'Front' face 1, perhaps naming it 'Face 1' or 'Front Face 1' and naming the next face 'Face 2' instead of 'Face 1', and so on.

perhaps!  I'm going to go with what we have for now to get this all moving, but i think we will revisit along these lines.  Thanks for the suggestion.

PR #888, opened 2025-04-14
Update nrdb.js

talha-ahsan:
25.04 allows for 4 5/3s so I added another check to support both in startup


plural:
You'll want to update asset_fingerprint in the app/config/parameters.yml file on prod as well.  You can use the value of `date +%s`

plural (approved):

PR #905, opened 2025-06-09
</> for base link and advancement cost

### LINKED ISSUE #878: Comparators for advancement requirement term (like `g>5`)
For example, you can write `v>3`, but not `g>5` (only `g:6|7|8|9`):

https://www.netrunnerdb.com/find/?q=g%3E5&sort=release-date&view=images&_locale=en

NoahTheDuke:
Fixes #878

Also adds `l` (base link), which already had support by @distributive in #637 but wasn't included in `canDoNumeric`.


plural (approved):
Nice.  Thanks, Noah!

PR #907, opened 2025-06-11
Correct Startup legality status on zoom view

lahlidahlia:
~~1. The app:legality:active-mwl command now also sets the latest startup MWL to active. This means that multiple MWL can be active at the same time. But only the latest standard and startup MWL should be active. As far as I can see, this is only enforced by the active-mwl command. The active-mwl command also first set every mwl to be inactive.
2. The SearchController uses this information to set the card's startup legality which is picked up by the zoom view.~~

Edited: 
Back to only one active MWL at a time. Added a simple startup legality check on the assumption that the latest startup MWL is active.

First commit :) feedback welcome. I don't really know the deployment process, but we will need to call app:legality:active-mwl once after this is committed.

Manual testing (cards and what the startup button shows in my sandbox):
Tributary: banned
Pharos: banned
R+: banned
Cleaver: banned
Creative Commission (previously banned and now legal): legal
Sable (not in startup): Not legal
Muslihat: Legal
Hedge Fund: Legal
Dorm computer: Not legal

### DIFF
--- a/src/AppBundle/Controller/SearchController.php
+++ b/src/AppBundle/Controller/SearchController.php
@@ -516,9 +516,7 @@ class SearchController extends Controller
                     }
                     $cardinfo['versions'] = [];
                     $standard_legal = "available";
-
-                    // Startup legality is currently hard-coded since the DB doesn't know anything about it.
-                    $startup_legal = false;
+                    $startup_legal = "unavailable";
 
                     $rotated_count = 0;
 
@@ -528,7 +526,7 @@ class SearchController extends Controller
                         // The 2 tutorial-only identity cards are invalid for startup and standard formats.
                         if ($v['code'] == '30077' || $v['code'] == '30076') {
                             $standard_legal = "unavailable";
-                            $startup_legal = false;
+                            $startup_legal = "unavailable";
                             continue;
                         }
                         // Draft and terminal directive campaign cards are not legal in standard.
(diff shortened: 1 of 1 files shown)


src/AppBundle/Controller/SearchController.php
```diff
@@ -518,7 +518,7 @@ public function displayAction(
                     $standard_legal = "available";
 
                     // Startup legality is currently hard-coded since the DB doesn't know anything about it.
```
talha-ahsan: we should probably remove this comment now :D

src/AppBundle/Command/LegalityActiveMwlCommand.php
```diff
@@ -45,19 +55,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                }
+                $mwl->setActive(true);
+                $standardDone = true;
+            } else if (!$startupDone && strpos($mwl->getCode(), "startup") !== false) {
+                if ($mwl->getActive()) {
+                    $output->writeln("Startup already active: " . $mwl->getName());
+                } else {
+                    $output->writeln($mwl->getName() . " set as ACTIVE for Startup");
+                }
+                $mwl->setActive(true);
```
plural: i don't have a ton of time to do more testing tonight, but i worry that having 2 marked as active is going to cause trouble.

Moving all of the banlist stuff to the client side is a heavy handed option, though.

Another option to consider is to add a method to get an active startup banlist and apply that in SearchController rather than having 2 marked active.

plural (changes requested):

src/AppBundle/Command/LegalityActiveMwlCommand.php
lahlidahlia: Yeah I think just having a method that retrieves the latest startup banlist is probably more sensible than trying to hunt down all logic that presumes there is only one active banlist. Especially if NRDBc isn't meant to stick around for too long.

src/AppBundle/Controller/SearchController.php
lahlidahlia: Thanks :)

plural (approved):
This is great, thanks!

Making some new helpers in the mwl service is probably the next place to go for updating the deckbuilder if you want to follow that thread.

PR #909, opened 2025-06-15
Deckbuilder: Fixed a litany of search issues.

lahlidahlia:
The deckbuilder searcher has been plagued with bugs for 9 years. Now, deckbuilder search functionality should match the site search bar.

The bulk of this fix is to change structure of the search query. We've been using forerunnerdb which shares significant syntax with mongodb. Previously OR (|) searches for integer conditions were not handled at all (with text queries implicitly handled by the $in operation). Now, the OR queries are properly handled by wrapping them with an $or operator. To handle multiple $or operator in the same query, the entire query is then wrapped by an $and operator.
Example, for o:2|3, the query that is made:

### o:2|3
Old:

```
{
    "cost": {
        "$eq": [
            2    <----- Incorrect, $eq does not accept an array of
                        values. Also, the | token was not parsed
                        correctly, and the 3 value was dropped.
        ]
    },
    ...
}
```

New:
```
{
    "$and": [
        {
            "$or": [
                {
                    "cost": {
                        "$eq": 2
                    }
                },
                {
                    "cost": {
                        "$eq": 3
                    }
                }
            ]
        }, ...
    ]
}
```

Also, a minor fix in the state machine to properly deliniate | tokens (line 187 token text-without-quotes) - exclude the | token from string match. This fixes a bug where x:text1|text2 would error out.

# Examples 
### o:2|3|4
Old:
image

New:
image

Note: Carnivore, Demolisher, Keiko was correctly included in the new search.

### x:"weapon"|"virus"
Old:
image

New:
image

Note: Old search breaks here and shows everything.

### x:weapon|virus
Old:
image

New:
image

Note: Old search breaks here and shows nothing.

### x:access x:"first time"
Old:
image

New:
image

Note: Old search incorrectly only queries for the last condition (x:"first time") and ignores the rest.

### o:2|3 p<2
Old:
image

New:
image

Note: Old search is missing everything that costs 3.

### DIFF
--- a/web/js/nrdb.smart_filter.js
+++ b/web/js/nrdb.smart_filter.js
@@ -2,16 +2,16 @@
   var SmartFilterQuery = [];
 
   smart_filter.get_query = function(FilterQuery) {
-        var query = _.extend(FilterQuery, SmartFilterQuery);
-        if (!SmartFilterQuery.text) {
-            delete query.text;
-        }
-        return query;
+      var query = SmartFilterQuery.slice();
+      query.push(FilterQuery);
+      // Wrap the query with an $and operator
+      query = {"$and" : query};
+      return query;
   };
 
   smart_filter.handler = function (value, callback) {
     var conditions = filterSyntax(value);
-    SmartFilterQuery = {};
+    SmartFilterQuery = [];
 
     for (var i = 0; i < conditions.length; i++) {
       var condition = conditions[i];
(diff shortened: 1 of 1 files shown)


web/js/nrdb.smart_filter.js
```diff
@@ -116,65 +126,65 @@
...
           query = RegExp.$3;
         } else {
           cond = [ "", ":" ];
         }
         etat = 2;
       } else {
-        if (   query.match(/^"([^"]*)"(.*)/) // jeton "texte libre entre guillements"
-          || query.match(/^([^\s]+)(.*)/) // jeton "texte autorisé sans guillements"
+        if (   query.match(/^"([^"]*)"(.*)/) // token "text with quotes"
+          || query.match(/^([^\s|]+)(.*)/) // token text-without-quotes
```
lahlidahlia, line 187: Pointing out this functional change: Exclude | token in this regex.

NoahTheDuke:
incredible! thank you for fixing this

PR #912, opened 2025-06-28
Fix inf costs on tournament sheets

migueldlr:


plural (approved):
Nice! Thanks for the speedy patch!

PR #918, opened 2025-07-13
Patch up the apply MWL command to be a bit simpler and actually fix the right decks.

### LINKED ISSUE #841: confusing deck legality
I find it very confusing that the "legality" list for decks seems to stop at different versions for different decks. 

It makes it seem like some decks aren't Standard legal when it's just that the legality version being shown at the top is out of date. For example [this deck](https://netrunnerdb.com/en/decklist/78c5ccd4-0d5e-405b-a07f-9ad9ae7e33d9/tv-teia-jinteki-fast-advance-) from 2023 is in fact Standard legal at this time, but at a glance it seems as if it weren't. 

As of this comment, we are on legality version `24.09`, but that deck's legality tags list stops at `23.09` for some reason, see this screenshot:

Screenshot 2024-09-07 at 11 11 03 PM

Moreover, it seems that due to the deck not being tagged with 24.xx legality versions, it doesn't show up on the deck search page under legality versions where that deck is in fact standard legal:

Screenshot 2024-09-07 at 11 13 51 PM

Apologies if this is a misunderstanding on my part about how the legality tags work.

plural:
Fixes #841 

In general this + an update to this task's invocation on the prod server should keep deck legality in sync with new ban lists.

### DIFF
--- a/src/AppBundle/Command/LegalityApplyMwlCommand.php
+++ b/src/AppBundle/Command/LegalityApplyMwlCommand.php
@@ -67,37 +67,46 @@ class LegalityApplyMwlCommand extends ContainerAwareCommand
             $output->writeln("<info>Applying legality for MWL $mwlCode...</info>");
 
             // Find all decklists that do not have a legality present for the given MWL.
-            $countDql = "SELECT COUNT(d)
-                FROM AppBundle:Decklist d
-                WHERE NOT EXISTS (
-                    SELECT l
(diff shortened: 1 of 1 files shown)


alexzhaohong:
I took a look at the commit. Thanks for the background explanation. 

and it looks like you're iterating on decklist ids rather than Decklist entities which is a nice lil change

PR #920, opened 2025-07-14
2 small improvements

lahlidahlia:
1) typeahead doesn't recognize Phat in zoom view searchbar. This is because lodash _.deburr doesn't correctly deburr Phat. Fix by using card's stripped_title instead.
Live:
image

Mine:
image

2) Don't populate zoom view searchbar with the card's code. This is exceptionally annoying because you have to backspace the code before being able to use the searchbar.
image

### DIFF
--- a/app/Resources/views/Search/searchbar.html.twig
+++ b/app/Resources/views/Search/searchbar.html.twig
@@ -26,11 +26,11 @@
                   return _.deburr(cardTitle).toLowerCase().trim();
                 }
                 var matchingCards = _.filter(latest_cards, function (card) {
-                    return regexp.test(normalizeTitle(card.title));
+                    return regexp.test(normalizeTitle(card.stripped_title));
                 });
                 matchingCards.sort((card1, card2) => {
-                    var card1title = normalizeTitle(card1.title);
-                    var card2title = normalizeTitle(card2.title);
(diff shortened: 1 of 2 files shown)


app/Resources/views/Search/searchbar.html.twig
```diff
@@ -62,7 +62,7 @@
     <div class="col-sm-4">
       <div class="input-group" style="margin-bottom:.5em">
         <input class="form-control" size="30" name="q" id="search-input"
-          tabindex="1" value="{{ q|default('') }}"
+          tabindex="1"
```
plural: how about specifying a default behavior that auto-selects the entire contents instead of clearing this out?

This also removes a search like `d:corp town` from the results which will be a pain for people iterating on their searches and I don't think people will like this behavior change.

plural (changes requested):

app/Resources/views/Search/searchbar.html.twig
lahlidahlia: Oh yeah... I'll take a closer look.

lahlidahlia: Ok, I'm now checking that the query matches the card code before omitting the query from the searchbar. In other cases the query will still be present after a search.

plural (approved):

PR #922, opened 2025-07-14
Sort by cost now sorts by cost first, then card type

### LINKED ISSUE #843: broken sort (bug)
**Issue:** 

This search is first sorting by card type, and then by card cost:
"Sort by Cost" & `d:r b:standard-ban-list-24-09 z:rotation-2023` 

Currently, if I go to the last page of the search results, this is what I see:

Screenshot 2024-09-10 at 11 01 08 AM

**Expected behavior:**

I was expecting to see `Lobisomem` and other cards in that last page, which are even more expensive. My assumption is that it first sorts by cost, and then by other things (like type).

lahlidahlia:
Fixes #843

Sort was unintuitively sorting by card types first before sorting by cost, which leads to confusion like above issue.

Live:
image

Mine:
image

A small side effect of this fix was that without specifying side corp and runner sides were all mixed up when using this sort option. So I also sorted by side first before sorting by cost. I don't know if it's more intuitive to sort sides first. I'd like others opinion on this.

Side note I double checked the other sort options and they all looked reasonable to me.

### DIFF
--- a/src/AppBundle/Service/CardsData.php
+++ b/src/AppBundle/Service/CardsData.php
@@ -802,7 +802,7 @@ class CardsData
                 $qb->orderBy('c.side', 'DESC')->addOrderBy('c.type')->addOrderBy('c.faction');
                 break;
             case 'cost':
-                $qb->orderBy('c.type')->addOrderBy('c.cost')->addOrderBy('c.advancementCost');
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
i love this

plural:
I think this works how people would expect now, so thank you!

PR #923, opened 2025-07-15
[PNP] Cut marks, clear button, importing deck sorts it by type

lahlidahlia:
1) Importing deck sorts the deck by type, prioritizing id - I think it's better user experience to see their ID appear first when importing their deck.
2) Non-invasive cut marks
image

3) Clear button. It also warns the user before clearing the list.
image


plural:
YESSSSSS this is great.

plural (approved):

PR #924, opened 2025-07-18
[PNP] PNP from public decklists, message change, small routing change

lahlidahlia:
PNP accessible through public decklists

image

Some routing changes, and change the alert to not suggest to users that they can print non-nsg cards through other websites (manveruppd mentioned we probably shouldn't say that).


app/Resources/views/Decklist/decklist.html.twig
```diff
@@ -80,6 +80,7 @@ NRDB.user.params.decklist_id = Decklist.id;
       <ul class="dropdown-menu" id="btn-group-decklist">
         <li><a href="#" id="btn-copy-decklist">Copy into my decks</a></li>
       <li><a href="#" id="btn-compare">Compare with another decklist</a></li>
+      <li><a href="#" id="btn-pnp">Print & Play</a></li>
```
plural: &amp;amp; instad of a raw & please

plural (approved):
just one thing.  This works really nicely!

Issue #903, opened 2025-05-11
Rotated IDs to bottom of faction pages


carlsondc:
Since about 2/3 of IDs have rotated at this point, having the standard legal ones at the top of the faction page would make it easier to browse for most users.

PR #926, opened 2025-07-22
Reorganise ID in faction page

mankinchi:
This is to handle this request https://github.com/NetrunnerDB/netrunnerdb/issues/903

RIght now the only order it has is alphabetically. In this PR, I have the active one on top, then banned, then rotated IDs.

I also clean up the code where it makes sense for contributors after me. This is my first PR so I might not be aware of all the team code contribution requirements. Please inform me if you need any changes ;)


src/AppBundle/Controller/FactionController.php
```diff
@@ -35,83 +36,77 @@ public function factionAction(string $faction_code, EntityManagerInterface $enti
...
+                    'isRotated' => $latestVersionOfIdentity->getPack()->getCycle()->getRotated(),
                     'decklists' => $list,
                 ];
             }
 
-            // Sort the identities alphabetically.
-            usort($decklists, function ($a, $b) {
+            // Banned identities go to the end
+            // then rotated then alphabetically
+            $getStatus = function($card) {
```
plural: This is throwing an error in my local dev setup using the docker config that is checked in.

I think you can rewrite this to move this to a private function in this class instead of a local variable.

if you make this a `private function` after this method you can call it with $this->getStatus instead of this local method.

I made that patch locally and it worked.

src/AppBundle/Controller/FactionController.php
```diff
@@ -35,83 +36,77 @@ public function factionAction(string $faction_code, EntityManagerInterface $enti
...
+                if ($statusA < $statusB) return -1;
+                if ($statusA > $statusB) return 1;
+
                 return strcasecmp($a['identity']->getTitle(), $b['identity']->getTitle());
             });
 
             $result[] = [
-                'faction'   => $faction,
-                'decklists' => $decklists,
+                'faction'         => $faction,
```
plural: remove the tab here.

src/AppBundle/Repository/MWLRepository.php
```diff
@@ -0,0 +1,30 @@
...
+    public function getBannedCardCodes() {
+        $activeMWLs = $this->findActiveMWLs();
+        $result = [];
+        foreach ($activeMWLs as $mwl) {
+            $result = array_merge($result, $mwl->getBannedCardCodes());
+        }
+
+        return $result;
+    }
+}
```
plural: please make sure your editor is adding newlines at the end of files.

plural (changes requested):
This is a nice improvement!

We may want to discuss aligning this more with the id selector controls on the create deck page, but i'll leave that up to you since that also handles multiple formats.

Image

src/AppBundle/Controller/FactionController.php
mankinchi: yeah, I was in thought of inline function from JS but PHP doesn't seem to have that. I have moved the function down to a private function

plural (approved):

Issue #915, opened 2025-07-12
Deck importer is unhelpful for cards with multiple printings


plural:
Which Hedge Fund would you like?  

* Hedge Fund
* Hedge Fund
* Hedge Fund
* Hedge Fund

<img width="758" height="331" alt="Image" src="https://github.com/user-attachments/assets/fbf370ba-b3b0-477f-b415-eff2765c7d1c" />

lahlidahlia:
I think I will eventually get around to refactoring the deck importer to use the same code the PNP is using. But PNP defaults to latest printing and show set names for each options.

PR #927, opened 2025-07-22
Add more details to options during deck direct import 

mankinchi:
This is to handle this issue https://github.com/NetrunnerDB/netrunnerdb/issues/915

If there is multiple printing, the option shows what set the card is from (it only change the dropdown, not the actual list)
image

Example in case of not similar text. If this is too much, please inform so I can modify it
image

### DIFF
--- a/web/js/directimport.js
+++ b/web/js/directimport.js
@@ -1,138 +1,128 @@
-$(document).on("data.app", function () {
-	$("#btn-import").prop("disabled", false);
-});
+$(document).on('data.app', function () {
+  $('#btn-import').prop('disabled', false)
+})
 
 $(function () {
-	$("#analyzed").on(
-		{
-			click: clickOption,
-		},
-		"ul.dropdown-menu a"
-	);
-	$("#analyzed").on(
-		{
-			click: click_trash,
-		},
-		"a.glyphicon-trash"
-	);
-});
+  $('#analyzed').on(
+    {
+      click: clickOption,
+    },
+    'ul.dropdown-menu a'
+  )
+  $('#analyzed').on(
+    {
+      click: click_trash,
+    },
+    'a.glyphicon-trash'
+  )
+})
 
 function click_trash(event) {
-	$(this).closest("li.list-group-item").remove();
-	updateStats();
+  $(this).closest('li.list-group-item').remove()
+  updateStats()
 }
 function do_import() {
-	$("#analyzed").empty();
-	const content = $('textarea[name="content"]').val();
-	const lines = content.split(/[\r\n]+/);
+  $('#analyzed').empty()
+  const content = $('textarea[name="content"]').val()
+  const lines = content.split(/[\r\n]+/)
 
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
This is killer!

plural:
Nice! Are you able to sort printings reverse chronologically?

mankinchi:
In term of release date? Should not be complicated

web/js/directimport.js
```diff
@@ -1,101 +1,109 @@
...
+			click: click_trash,
+		},
+		"a.glyphicon-trash"
+	);
 });
 
 function click_trash(event) {
-  $(this).closest('li.list-group-item').remove();
-  update_stats();
+	$(this).closest("li.list-group-item").remove();
```
plural: while we don't have a linter set up for classic, please use spaces instead of tabs to stay consistent with the rest of the codebase.

plural (changes requested):

web/js/directimport.js
```diff
@@ -55,6 +55,31 @@ function getCardRowForLine(line, lineNumber) {
...
+				// Keep cards with the same name as the first card at the top
+				const aHasSameTitle = a.title === cards[0].title;
+				const bHasSameTitle = b.title === cards[0].title;
+
+				if (aHasSameTitle && !bHasSameTitle) return -1;
+				if (!aHasSameTitle && bHasSameTitle) return 1;
+
+				// If same title, sort by pack release date reversed
+				if (a.title === b.title) {
+					const aPackReleaseDate = moment(
```
plural: i don't think you need the moment creation.  The date format is already sortable properly as a string.

plural (changes requested):

web/js/directimport.js
```diff
@@ -1,101 +1,128 @@
...
+            // Keep cards with the same name as the first card at the top
+            const aHasSameTitle = a.title === cards[0].title
+            const bHasSameTitle = b.title === cards[0].title
+
+            if (aHasSameTitle && !bHasSameTitle) return -1
+            if (!aHasSameTitle && bHasSameTitle) return 1
+
+            // If same title, sort by pack release date reversed
+            if (a.title === b.title) {
+              return b.pack.date_release.localCompare(a.pack.date_release)
```
plural, line 67: this should be localeCompare.

web/js/directimport.js
plural, line 1: please adjust the formatter settings you set up.  It removed all trailing semicolons, which is a big, unnecessary change and will update a ton of files.

I would recommend not doing the linter / formatter at all in this PR, but as it's own PR so we can evaluate it's settings and impact.

plural (changes requested):

web/js/directimport.js
mankinchi: oh, I didn't realise that it removes the trailing semicolons (which I also don't like). I have prettier installed so it runs by default. I'll revert those changes 🐱

PR #934, opened 2025-08-17
Add Eternal Points to individual card pages

talha-ahsan:
Right now we have Standard Card Pool and Standard Ban List views. However the only case where those would differ is when the card is still banned but rotated (because I forgot to remove it from the ban list). Consolidating those together into "Standard Card Pool" while retaining the historical view should avoid any loss of information. 

In the place of "Standard Ban List" there's a new entry that tracks Eternal Points costs. This hits the new API, finds the points value, and updates the cell. 

Right now it's a bit sluggish because we need a new API call so there's clear client side updating for the eternal points, but otherwise I think this is good to PR?

image

 versus Watch the World Burn which is still banned 
 
 
image

Also I partially refactored the card pools section to not be so heavily based on a table. There's client side JS in zoom.js that needs to be updated before I can pull all of the table based layout from this section.

### DIFF
--- a/app/Resources/views/Search/display-zoom.html.twig
+++ b/app/Resources/views/Search/display-zoom.html.twig
@@ -321,16 +321,16 @@
                                   {% endif %}
                                 {% endfor %}
                             > Standard Card Pool</th>
+                            <a id="mwl-history" href="">(show history)</a>
                         </tr>
                         <tr>
-                            <th>
-                                <span
+                            <th
                                   {% for mwl_info in card.mwl_info %}
                                     {% if mwl_info.active %}
                                       class="legality-{{ mwl_info.legality }}"
                                     {% endif %}
-                                  {% endfor %}> Standard Ban List</span>
-                                  <a id="mwl-history" href="">(show history)</a>
+                                  {% endfor %}> Eternal Card Pool
+                                  
                             </th>
                         </tr>
                     </thead>
--- a/app/Resources/views/Search/display-zoom.html.twig
+++ b/app/Resources/views/Search/display-zoom.html.twig
@@ -42,6 +42,31 @@
         });
     }
 
+    async function retrieveEternalPoints() {
+        // retrieves the latest Eternal points list 
+        // and then checks if the card is banned, or in any of the points fields 
+        // if not, it's worth 0 points. 
+        const formatInfo = await Promise.all(
+            fetchFullDataResponse(`${v3_api_url}/api/public/formats/eternal`)
+        );
+        // the API sometimes is out of date for active_card_pool_id 
+        const latest_points_code = formatInfo.data.attributes.restriction_ids.at(-1);
+        const [pointsData, printing] = await Promise.all( 
+            fetchFullDataResponse(`${v3_api_url}/api/public/restrictions/${latest_points_code}`),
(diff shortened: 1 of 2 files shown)


app/Resources/views/Search/display-zoom.html.twig
```diff
@@ -42,6 +42,28 @@
...
 
+    async function retrieveEternalPoints() {
+        // retrieves the latest Eternal points list 
+        // and then checks if the card is banned, or in any of the points fields 
+        // if not, it's worth 0 points. 
+        const formatInfo = await fetchFullDataResponse(`${v3_api_url}/api/v3/public/formats/eternal`);
+        const latest_points_code = formatInfo.data.attributes.active_restriction_id;
+        const [pointsData, printing] = await Promise.all( 
+            [
+                fetchFullDataResponse(`${v3_api_url}/api/v3/public/restrictions/${latest_points_code}`),
```
plural: you may be able to pull this down with the printing info, but i would not hold this up for that.  

we need to check out our caching as well, which will minimize the hit if we are doing it correctly.

web/css/style.css
```diff
@@ -145,6 +146,8 @@
   --nrdb-color--banned: #7F4554;
   --nrdb-color--rotated: #37435E;
   --nrdb-color--restricted: #674B76;
+    --nrdb-color--eternal-points: #B85C00;
```
plural: nit: alignment

web/css/style.css
```diff
@@ -1635,6 +1643,32 @@ ul.rulings-list blockquote {
...
+  content: "2 POINTS";
+  background-color: var(--nrdb-color--eternal-points);
+}
+
+.legality-eternal-3-points:before {
+  content: "3 POINTS";
+  background-color: var(--nrdb-color--eternal-points);
+}
+
+
```
plural: nit: drop extra blank line

plural (approved):
2 nits and one thing we can follow up on separately.

app/Resources/views/Search/display-zoom.html.twig
lahlidahlia: Oh yeah, looks like you can. `printing.data.attributes.restrictions.points`

Additionally, I think you should refactor and bandwagon this API call with the one used for flip ids and narrative backs (line ~71). This should speed things up a lil.

lahlidahlia:
One suggestion, do you think it would be a good idea to put a placeholder like this 

image

before we finish gathering eternal information so that the text doesn't look like it's bouncing around as it loads in the eternal points?

lahlidahlia (changes requested):

web/css/style.css
```diff
@@ -1635,6 +1643,32 @@ ul.rulings-list blockquote {
...
+.legality-eternal-4-points:before {
+  content: "4 POINTS";
+  background-color: var(--nrdb-color--eternal-points);
+}
```
lahlidahlia, line 1675: Not a showstopper, but for future proofing you might want to also add 5, 6 and 7 points too so that if they ever decided to update the points list with those numbers we won't have to submit another PR for it and it would just work.

talha-ahsan: so for this I ended up reaching out to the eternal folks, and there's no plan at the moment to go beyond 4 points, hence the discussion to not do 5

lahlidahlia (approved):
LGTM!

PR #936, opened 2025-08-19
Use xlarge webp for zoom view and PNP, add PNP progress bar

lahlidahlia:
The images, they're crispy.

This depends on https://github.com/NetrunnerDB/netrunnerdb-api-server/pull/411, but is still able to appropriately fallback to large jpgs without it.

An unintended consequence of using xlarge webps is that a fresh PNP takes ~17 seconds and that's honestly way too long. I don't really know how I could speed this up but I can at least make this more palatable by adding a progress bar. ~This might involve moving the pdf generation part to a web worker. I'd like to do this before all this goes live if possible.~

I've also refactored the print code to not be so UI blocking and added a progress bar. Print is still slow but it's honestly pretty palatable with a nice progress bar.
image

### DIFF
--- a/app/Resources/views/Search/display-zoom.html.twig
+++ b/app/Resources/views/Search/display-zoom.html.twig
@@ -78,7 +78,12 @@
                 button.innerHTML = buttonText;
                 button.dataset.face = i;
                 // Use the main image for the first button and the face image for i - 1 (because we are combining the front face with the faces array)
-                let buttonImg = i == 0 ? printing.data.attributes.images.nrdb_classic.large : printing.data.attributes.faces[i-1].images.nrdb_classic.large;
+                let buttonImg = null;
+                if(printing.data.attributes.images.nrdb_classic.xlarge) {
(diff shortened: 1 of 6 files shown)


app/Resources/views/Search/display-zoom.html.twig
```diff
@@ -78,7 +78,12 @@
                 button.innerHTML = buttonText;
                 button.dataset.face = i;
                 // Use the main image for the first button and the face image for i - 1 (because we are combining the front face with the faces array)
-                let buttonImg = i == 0 ? printing.data.attributes.images.nrdb_classic.large : printing.data.attributes.faces[i-1].images.nrdb_classic.large;
+                let buttonImg = null;
+                if(printing.data.attributes.images.nrdb_classic.xlarge) {
```
plural: nit: add space after `if`

plural: (throughout the PR)

plural (approved):
couple of nits, but this is nice!

PR #952, opened 2026-01-07
Add endpoints for keycloak user migration.

plural:
This new endpoint (get and post versions) support migrating users into Keycloak using this method: https://codesoapbox.dev/keycloak-user-migration/

### DIFF
--- a/src/AppBundle/Controller/UserMigrationController.php
+++ b/src/AppBundle/Controller/UserMigrationController.php
@@ -1,0 +1,133 @@
+<?php
+
+namespace AppBundle\Controller;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Psr\Log\LoggerInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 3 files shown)


src/AppBundle/Controller/UserMigrationController.php
```diff
@@ -0,0 +1,133 @@
...
+            'id' => null,
+            'username' => $user->getUsername(),
+            'email' => $user->getEmail(),
+            // NRDB Classic does not have first/last name fields
+            'firstName' => null,
+            'lastName' => null,
+            'enabled' => $user->isEnabled(),
+            'emailVerified' => false,
+            'attributes' => [
+                'faction' => $user->getFaction(),
```
rlankin, line 66: Is this the username faction color from the user's profile? If so, should any of the other settings be included here?

rlankin (approved):

src/AppBundle/Controller/UserMigrationController.php
plural: yeah, it is the faction selection from there.  i picked this one for 2 reasons.  1 - i would actually like to have the keycloak profile let people make lovely custom id cards.  2 - i forgot there were other settings.  :)

We may not even keep this bit, so let's revisit what settings to move into the ID vs. to move into NRDB v2 later.

PR #971, opened 2026-02-23
PNP: Melies U and non-owner deck view PNP button

lahlidahlia:
### Melies U
Hard-coded because I think it's not worth making expensive(?) api call for every card added to PNP. Hopefully we only have to continue doing this until V2.

image

### Non-owner deck view PNP button
I would appreciate a review on the permission side of things. The PNP deck endpoint now only checks if the user is deck owner or deck owner has share deck enabled (previously required user to owner to use the PNP deck endpoint).

image


talha-ahsan (approved):

web/js/pnp.js
plural, line 13: i'll file a bug for num_faces not being exposed yet as a search operator.

https://api.netrunnerdb.com/api/v3/public/cards?filter[search]=designed_by:null_signal_games%20t:corp_identity|runner_identity

release_date, date_release, r should all let you search by a date filter as well.

PR #972, opened 2026-02-25
Add skeleton of V3 search UI.

plural:
New client side only search page for the new search endpoint.

it is a fun, zippy experience!

Adds syntax page for v3 and navigation integration as well.

### DIFF
--- a/src/AppBundle/Controller/SearchNewController.php
+++ b/src/AppBundle/Controller/SearchNewController.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace AppBundle\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+
+class SearchNewController extends Controller
+{
+    /**
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function getAction(Request $request)
+    {
+        return $this->render('/search/search_new.html.twig', [
+            'pagetitle'       => "Card Search (new)",
+            'pagedescription' => "New Card Search",
+            'query'           => $request->query->get('q') ?: ''
+        ]);
+    }
+}
--- a/src/AppBundle/Controller/SearchNewController.php
+++ b/src/AppBundle/Controller/SearchNewController.php
@@ -15,7 +15,9 @@ class SearchNewController extends Controller
         return $this->render('/search/search_new.html.twig', [
             'pagetitle'       => "Card Search (new)",
             'pagedescription' => "New Card Search",
-            'query'           => $request->query->get('q') ?: ''
+            'query'           => $request->query->get('q') ?: '',
+            'sort'            => $request->query->get('sort') ?: '',
+            'view'            => $request->query->get('view') ?: '',
         ]);
     }
 }
--- a/src/AppBundle/Controller/SearchNewController.php
+++ b/src/AppBundle/Controller/SearchNewController.php
@@ -13,8 +13,9 @@ class SearchNewController extends Controller
     public function getAction(Request $request)
     {
         return $this->render('/search/search_new.html.twig', [
-            'pagetitle'       => "Card Search (new)",
             'pagedescription' => "New Card Search",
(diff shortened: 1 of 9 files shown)


src/AppBundle/Controller/SearchNewController.php
```diff
@@ -0,0 +1,21 @@
...
+use Symfony\Component\HttpFoundation\Request;
+
+class SearchNewController extends Controller
+{
+    /**
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function getAction(Request $request)
+    {
+        return $this->render('/search/search_new.html.twig', [
```
lahlidahlia: /search/search_new... -> /Search/search_new...

lahlidahlia:
Thank you for getting started on this!

Some thoughts as I am perusing this and thinking about it:
- I think the new syntax is pretty overwhelming and is a learning curve for regular users for sure (me included). I think at a minimum we'll need to coopt the syntax guide page to try and teach users the syntax. I'll take on the work for this. It'll help me get my feet wet on the v3 search syntax as well and look for any opportunities for improvement.
- It would also be nice if we have a syntax autocompletion, similar to @migueldlr's app that he shared some time ago. This would help with syntax learning since some of the keyword seems verbose or it could prompt user on shortcodes.
- Rather than an account toggle, I think we should do a toggle on the search bar, so that it's more visible, and so user can rapidly switch back and forth between new and old search as they like.

plural:
> Thank you for getting started on this!
> 
> Some thoughts as I am perusing this and thinking about it:
> 
> * I think the new syntax is pretty overwhelming and is a learning curve for regular users for sure (me included). I think at a minimum we'll need to coopt the syntax guide page to try and teach users the syntax. I'll take on the work for this. It'll help me get my feet wet on the v3 search syntax as well and look for any opportunities for improvement.

Absolutely.  I was thinking about a table with v2 on the left v3 on the right to start for anything with updated keywords AND to help us tidy up any names we might have missed.  Additionally, we can add the new functionality more clearly as well. https://api.netrunnerdb.com/api/docs/cards%2Ffilter_-_card_search_operator is a good starting point for this.  Do you want to start there by maybe updating the syntax page or taking a crack at it somewhere? 

> * It would also be nice if we have a syntax autocompletion, similar to @migueldlr's app that he shared some time ago. This would help with syntax learning since some of the keyword seems verbose or it could prompt user on shortcodes.

Absolutely.  that's more "flesh" than "bone" though, so i am trying to get the bones all in place that we can build on here.

> * Rather than an account toggle, I think we should do a toggle on the search bar, so that it's more visible, and so user can rapidly switch back and forth between new and old search as they like.

I was certainly planning on updating the nav to offer clear links to both and think we should have both an account setting (done!) and a local toggle.

plural:
Fixes #978

lahlidahlia:
> Absolutely. I was thinking about a table with v2 on the left v3 on the right to start

I was thinking that we'd have two separate pages that can be accessed through "syntax" on the navbar, defaulting to whichever is selected in your profile setting, with a button on top of each to switch to the other one. I think we can't really cram both pages worth of instruction into one that's probably too much, and I like the current layout of the syntax page as is. Also this way it's easy to copy the syntax page to a new one without much work. I'll get started on this soon.

NoahTheDuke:
if i'm reading the twig right, this is using the url query syntax directly as the query syntax that folks will use? ~~how does the underscore in `_:/^b/ f:anarch t:program` work?~~ jk i forgot this already exists lol.

plural:
> if i'm reading the twig right, this is using the url query syntax directly as the query syntax that folks will use? ~how does the underscore in `_:/^b/ f:anarch t:program` work?~ jk i forgot this already exists lol.

`_` works the same way it always as - as a synonym for card name/title:  https://netrunnerdb.com/find/?q=_%3Achange

plural:
I mean, you can be forgiven for finding out things that have existed in NRDB classic for a long time, @NoahTheDuke .  I'm still surprised by something most weeks!

PR #984, opened 2026-03-02
added VP to startup, removed Liberation

talha-ahsan:


src/AppBundle/Controller/SearchController.php
```diff
@@ -562,7 +562,7 @@ public function displayAction(
                     // officially released, the card is not standard legal.
                     if ($rotated_count == count($cardinfo['versions']) ||
                         $pack->getDateRelease() == null ||
-                        $pack->getDateRelease()->format("Y-m-d") >= date("Y-m-d")) {
```
plural, line 565: Why make this chnage?

plural (approved):

talha-ahsan:
Currently on the card page, vantage point cards don’t show as legal even though as of  today they officially released. It feels weird to say cards are only legal the day after they are released.

-
________________________________

PR #990, opened 2026-03-14
Add new jnet deck meta-data when exporting deck

jwarwick:
JNet will have support for new meta-data when copying in a deck list that allows setting the identity and deck name. Added those fields to the jnet export function.

Jnet PR here: https://github.com/mtgred/netrunner/pull/8611

### DIFF
--- a/web/js/nrdb.js
+++ b/web/js/nrdb.js
@@ -1144,6 +1144,14 @@ function build_jintekinet(deck) {
     var deck = process_deck_by_type(deck || SelectedDeck);
     var lines = [];
 
+    lines.push(";; identity: " + Identity.title);
+    lines.push(";; title: " + SelectedDeck.name);
+    var deckUrl = (typeof Decklist != "undefined" && Decklist != null)
+        ? location.origin + Routing.generate('decklist_view', {decklist_uuid: Decklist.uuid})
+        : location.origin + Routing.generate('deck_view', {deck_uuid: SelectedDeck.uuid});
+    lines.push(";; notes: " + deckUrl);
+    lines.push("");
+
     $('#deck-content > div > div > div').each(function (i, line) {
         var num = $(line).contents().filter(function () {
             return this.nodeType == 3;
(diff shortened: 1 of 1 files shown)


web/js/nrdb.js
```diff
@@ -1144,6 +1144,14 @@ function build_jintekinet(deck) {
     var deck = process_deck_by_type(deck || SelectedDeck);
     var lines = [];
 
+    lines.push(";; identity: " + Identity.title);
```
plural: are you sure you want to do title matching vs. either a printing code or a (new) id  value?

plural: or maybe have the code/id additionallly?

NoahTheDuke: Are there issues with using the title? Do you expect there to be overlap or duplicates?

plural:
We have already caused problems where people are matching on text (hi nice
quotes) and card corrections do happen (although mostly rules text). The
codes and ids are stable, though

-jason

NoahTheDuke:
Given that nrdb is generating this and jnet is ingesting it (with a person merely copy-pasting between the two), do we need to worry about that? The deck list doesn't use ids or codes, it uses card titles and that's been fine for years. What is special about Identities?

jwarwick:
I think it's pretty low risk. Everything else is just the text name and I do run it through a bit of a fuzzy search when we parse on the other end. Worst case the user just needs to pick from the ID selector in the deckbuilder like they do today. I don't think it's worth overthinking.

What I really should do is revive the ability to access your NRDB decks from within jnet, but that's a whole different project.

plural:
Agreed it is low risk, but i am still smarting from the ABR issues caused
by text matching ids.


We should talk decks and stuff later. We have a brand new deck validation
api and some other stuff cooking

-jason

NoahTheDuke:
except for playtest cards, we rely on nrdb for our cards data so any special characters should match exactly. people hand writing stuff might get bad or incorrect data but like john says, they can manually fix it.

jwarwick:
If we're really concerned, I can also add parsing for `identity-card-num` and we can populate it here. I don't have particularly strong feelings either way.

plural (approved):
sorry to keep you hanging - i've been swamped.  We'll get this pushed by the weekend.

PR #992, opened 2026-03-18
Simplify Docker setup

alsciende:
Hello folks. I wanted to set up a local nrdb and found your cool Docker files. A few things in the docker-compose.yaml file surprised me and I tried to understand it. I think I found a simpler setup, let me know what you think. It works for me.

- removed the named volumes on /vendor and /var which are in .gitignore anyway
- removed the named volumes on /bin and /web because there's no reason they would be changed unexpectedly. The data in /web is already ignored by .gitignore.
- removed the named volumes on /app/config because I created a mount volume on /app/config/parameters.yml instead of copying files around

Consequently, I simplified `docker/docker-first-run.sh` too because no files need to be copied.

Benefits of this MR:

- Faster setup
- Easier to understand and maintain
- The vendors installed by "composer install" in vendor/ are accessible outside the container, so available to the IDE
- The logs files in var/log/ are accessible outside the container, should you need them

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -22,6 +22,7 @@ nrdb-dev-env.
 cd ..
 git checkout https://github.com/netrunnerdb/netrunnerdb
 git checkout https://github.com/netrunnerdb/netrunner-cards-json
+cp -r netrunnerdb/docker nrdb-dev-env
 cd nrdb-dev-env
 ```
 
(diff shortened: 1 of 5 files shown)


plural (approved):
looks great, thanks!  We set up the initial docker stuff a long time ago and the two of us that did it were fairly new to docker so it was ... messy!

PR #1002, opened 2026-04-06
Separate influence by quantity instead of always 5 pips

cls:
I expect this has the potential be contentious, but I would find it much easier to see at a glance how much influence is being spent and how much could be saved by cutting a card, for example, from "•• •• ••" or "••• •••" than "••••• •" as would currently be displayed. I don't do PHP, but I've done my best to suggest a change.


cls:
Oh, obviously this should be `influence / qty` rather than just `qty`.

PR #1013, opened 2026-07-24
Quick patch to sorting behavior for cost.

plural:


app/Resources/views/Search/search_new.html.twig
```diff
@@ -267,6 +267,18 @@ async function fetchResults(queryUrl) {
 
 function sortResults() {
   // Sort results (this needs to refactored a bit so sorting can be done without refetching data).
+  function numForSorting(s) {
+    return s === 'X' ? -1 : (s !== '' && s !== null && s !== undefined && !isNaN(Number(s))) ? Number(s) : -100;
```
migueldlr: This ternary is a bit hard to read

app/Resources/views/Search/search_new.html.twig
```diff
@@ -277,6 +289,16 @@ function sortResults() {
       } else {
         return 0;
       }
+    } else if (sortBy == 'cost') {
+      const aVal = numForSorting(getCostVal(a));
```
migueldlr: It looks like `numForSorting` and `getCostVal` are only used in these two places and always used together - could we combine the two methods into one?

migueldlr (changes requested):

app/Resources/views/Search/search_new.html.twig
plural: Yep.  Extracted this out into a new helper.

app/Resources/views/Search/search_new.html.twig
plural: yep.  Combined here and did the re-use up in the local helper.

migueldlr (approved):

